PR #6, opened 2023-01-31
Require build-ui before building Go binary.

RobbieMcKinstry:
The Go binary embeds the web UI at compile time. For `make bin` to succeed, the web UI must already have been build. As a result, a fresh install of the repo fails with `make dev`. With this change, the web UI is always rebuilt before the Go binary is built, ensuring the Go server has the latest version of the web UI. This fixes `make dev`.


anubhavmishra:
The `make build` command now builds the UI and the Go binary. `bin` target only builds the binary by design. Thanks for the interest in Hermes!

PR #7, opened 2023-02-01
Add "Subscribed" and "Removed" messages to switches; Fix broken tests

jeffdaley:
Confirms user actions by adding a temporary confirmation message onClick.
In addition, fixes tests that broke when we switched from mock data to database-stored information.

https://user-images.githubusercontent.com/754957/215927183-ea80c4fc-d856-4418-80d9-fd3d583ba4f7.mp4


jfreda (approved):
Looks good!

PR #18, opened 2023-02-02
Add SessionService types

jeffdaley:
Adds types to the `ember-simple-auth` SessionService and replaces uses of `session: any`.


jeffdaley:
Closing and wrapping this into #20

PR #22, opened 2023-02-03
Initial work for CI

anubhavmishra:
PR adds an initial GH Action workflow for CI.


web/yarn.lock
```diff
@@ -10701,6 +10701,45 @@ __metadata:
   languageName: node
   linkType: hard
 
+"heimdalljs-fs-monitor@npm:^1.1.0":
```
anubhavmishra, line 10704: yarn lock file needs updating for the UI build. @jeffdaley could you please take a look why these changes in the lock file were triggered?

jeffdaley: I think it's just an alphabetizing change. Dunno how "heimdall" ever placed after "hermes" though...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches: [main]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: "read"
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
aldy505, line 16: should probably use v3 as v2 uses node 14, and v3 uses node 16

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: ci
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
```
aldy505, line 7: I'd prefer it targets every branch, so everyone gets the benefits of build and test commands. If there are other needs for packaging the source code or perform any other specific command that should be run on PR only, or on main branch only, we can create a separate workflow file for that.

```suggestion
on:
  pull_request:
  push:
```

.github/workflows/ci.yml
anubhavmishra: Good catch!

.github/workflows/ci.yml
```diff
@@ -21,19 +21,19 @@ jobs:
           node-version: "16"
           cache-dependency-path: "web/yarn.lock"
 
-      - name: yarn set version 3.3.0
-        run: yarn set version 3.3.0
+      - name: yarn set version
+        run: make web/set-yarn-version
```
anubhavmishra, line 25: Making this a make target so we can track version changes in one file.

Makefile
```diff
@@ -76,3 +81,7 @@ web/run: ## Run web application while proxying backend requests
 web/run: web/install-deps
 	cd web \
 		&& yarn start:with-proxy
+
+web/set-yarn-version: ## Set yarn version
```
jfreda, line 85: Nit: could use a `.PHONY` here too.

```suggestion
.PHONY: web/set-yarn-version
web/set-yarn-version: ## Set yarn version
```

jfreda (approved):

Makefile
anubhavmishra: Good catch!

PR #30, opened 2023-02-05
Update README to include renaming credentials

jrschumacher:
When following the README the build will fail unless the downloaded Google OAuth credentials file is renamed


anubhavmishra (approved):
@jrschumacher thanks for the fix!

PR #31, opened 2023-02-05
Add more detail around Algolia

jrschumacher:
When following the README I attempted to start the application without Algolia. When I went back and updated it was a bit unclear where the configuration was needed since there are three areas where config is added:

1. At build time via ENV for web app
2. At runtime via the Google credentials.json file
3. At runtime via the config.hcl file


jeffdaley (approved):

jeffdaley:
Good call @jrschumacher

Thanks for your contribution!

PR #32, opened 2023-02-06
show document title in the HTML page title

meirish:
I noticed with multiple tabs open, I was only ever getting "Document" in the page title. This should show the document's actual title instead.


jeffdaley (approved):

jeffdaley:
Wonderful, thank you. I'll soon write tests to catch regressions like this!

PR #37, opened 2023-02-08
Show active filters; Toolbar refactor

jeffdaley:
Shows which filters are active for a given route.

CleanShot 2023-02-08 at 14 11 20

The newly created `ActiveFilters` service captures a route's queryParams (i.e., active filters) whenever its model is loaded or refreshed. This creates a shared index of active filters for components and helpers to compute off. In general, this lets us be more efficient with queryParams: You'll notice the URLs are more concise, for example.

You'll also see that the facet and sort controls have switched from `button`s to `LinkTo`s. Instead of computing the queryParams `onClick`, we do it beforehand, which means we can pass them directly to the component's `@query` argument and our `onClick` function only has to close the dropdown.


web/app/components/header/facet-dropdown.ts
```diff
@@ -1,47 +1,38 @@
...
 interface FacetDropdownComponentSignature {
   Args: {
-    onClick: (facetName: FacetNames, value: string) => void;
     label: string;
     facets: FacetDropdownObjects;
     disabled: boolean;
   };
 }
 
-export enum FacetNames {
```
jeffdaley, line 14: Moved this to `toolbar.ts`

web/app/components/header/facet-dropdown.ts
```diff
@@ -1,47 +1,38 @@
...
-  DocType = "docType",
-  Owners = "owners",
-  Status = "status",
-  Product = "product",
-}
```
jeffdaley, line 19: Moved this to `toolbar.ts`

web/app/components/header/toolbar.js
```diff
@@ -1,121 +0,0 @@
...
-    this.toolbar.sortBy = value;
-
-    this.router.transitionTo({
-      queryParams: {
-        sortBy: value,
-      },
-    });
-    closeDropdown();
-  }
-}
```
jeffdaley, line 121: This is the removed chunk

jfreda (approved):
Looks reasonable to me - excited for this!

Design q (not blocking, just something I was thinking about): do you think we might want to differentiate the type of active filters somehow? It's _probably_ mostly obvious which type each one is, but wondering if we'd want to label them something like `type:RFC`, `product:Terraform` etc.  Not totally sure how I feel about this right now (there's a tradeoff of it looking busier) but wanted to see what you think.

jeffdaley:
@jfreda I tried it early on but it felt redundant, like the filters were obvious enough as types/products, etc. But I'll explore it further; it has potential to be helpful.

PR #38, opened 2023-02-09
Include Drafts in Recently Viewed Documents

jeffdaley:
Moves RecentlyViewedDocument functions from the dashboard route to a service. By detaching from the model hook, we make fewer network calls and get a faster, more reactive dashboard.

The service also expands to include drafts, and to accommodate legacy users, we transform the old list:

```
files: ['docID1', 'docID2', 'docID3']
```
into the new format when `fetchAll` is called by the dashboard route:
```
files:[
    { id: 'docID1', isDraft: false },
    { id: 'docID2', isDraft: false },
    { id: 'docID3', isDraft: false }
]
```

You can see for yourself locally. Log in and your recently viewed docs should be preserved. (Note: Your recent-docs list will reset when switching back to main.)


web/app/routes/authenticated/dashboard.js
```diff
@@ -67,52 +67,12 @@ export default class DashboardRoute extends Route {
...
-          )}`;
-
-          recentlyViewedDocs.push(promise.value);
-        }
-      });
-
-      return recentlyViewedDocs;
-    });
+    if (this.recentDocs.all === null) {
+      await this.recentDocs.fetchAll.perform();
```
jfreda, line 71: Should we add some error checking here?

web/app/routes/authenticated/document.js
```diff
@@ -103,10 +103,8 @@ export default class DocumentRoute extends Route {
       console.log("Error recording analytics: " + err);
     }
 
-    // If not a draft, record the doc as recently viewed.
-    if (!params.draft) {
-      this.recentDocs.addDoc.perform(params.document_id);
-    }
+    // Record the doc with the RecentlyViewedDocs service.
+    void this.recentDocs.markViewed.perform(params.document_id, params.draft);
```
jfreda, line 107: Should we add some error checking here? Curious about the use of `void` as well.

jfreda (commented):
Overall looks really good! Just wondering about error checking in a couple of spots so the dashboard doesn't break in case these new service methods error out.

web/app/routes/authenticated/dashboard.js
jeffdaley: What did you have in mind? Here I'm relying on `fetchAll`'s error handling

web/app/routes/authenticated/document.js
jeffdaley: Another place where I'm relying on `markViewed`'s error handlers. And because I'm not awaiting this task, I'm not sure what error handling I could do. But let me know your thoughts.

I use `void` on async functions (that return void) to signal when I'm _deliberately not awaiting them_. Here I just want to kick off the task and not have it hold up the `model` call. I'm hoping `void` reminds me of this the next time I'm in the code. Shouldn't have any affect on the function when used this way.

web/app/routes/authenticated/dashboard.js
jfreda: It looks like `fetchAll` just logs and throws an error so it would end up here? I might be missing something though.

web/app/routes/authenticated/document.js
jfreda: Okay, makes sense 👍

web/app/routes/authenticated/dashboard.js
jfreda: Also this isn't necessarily a blocker, but I just wanted to make sure that the dashboard would still load if there was an error with `fetchAll` - did you test that?

jfreda (approved):
Nice, just one sorta open question about error handling, but otherwise LGTM!

web/app/routes/authenticated/dashboard.js
jeffdaley: I see what you mean. Check out these changes: https://github.com/hashicorp-forge/hermes/pull/38/commits/a634c7d2d837e77620d2dc15cfa0632e9f87176e

We now let failed `fetchAll` through and show an error in the template with the option to refresh.

In this example it's hard-coded to fail, but it should give you an idea of the implementation:
https://user-images.githubusercontent.com/754957/227072262-c2cbc23e-28d7-4200-9186-2a3e8e4063cd.mp4

jfreda: Looks great! :shipit:

Issue #8, opened 2023-02-01
Add Docker image


inductor:
For someone who just want to try Hermes real quick I think it's nice to have a Docker image officially.

anubhavmishra:
Thanks for the interest in Hermes! This is on our short-term roadmap as part of some build and release improvements that we were not able to include as part of our initial release. We're attaching the planned-work label to this as something that HashiCorp Labs is going to do.

trungdlp-wolffun:
You can check my PR @inductor https://github.com/hashicorp-forge/hermes/pull/42

PR #42, opened 2023-02-10
Create dockerfile to build image

trungdlp-wolffun:
Issue: https://github.com/hashicorp-forge/hermes/issues/8


.dockerignore
```diff
@@ -0,0 +1 @@
+**/node_modules
```
inductor, line 1: Add EOF in the end

Dockerfile
```diff
@@ -0,0 +1,37 @@
+# BUILD FRONTEND STAGE
```
inductor, line 1:
```suggestion
# syntax = docker/dockerfile:1.4

# BUILD FRONTEND STAGE
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+FROM node:18-alpine as frontend-builder
```
inductor, line 2:
```suggestion
FROM node:16-alpine as frontend-builder
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Copy execute file
+WORKDIR /app
+COPY --from=backend-builder /app/hermes hermes
+
+# Set execuateable
+RUN chmod +x hermes
+
+# Execute app
+ENTRYPOINT ["./hermes"]
```
inductor, line 37: Add EOF in the end

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# Install Make
+RUN apk add --no-cache make
+
+# Install Yarn v3
+RUN corepack enable
+RUN corepack prepare yarn@stable --activate
+RUN yarn set version stable
+
+# Build frontend
+RUN make web/build
```
inductor, line 17:
```suggestion
WORKDIR /app

# Install Make
RUN apk add --no-cache make

# Install Yarn v3
RUN corepack enable
RUN corepack prepare yarn@stable --activate
RUN yarn set version stable

# Copy and install deps
COPY --link Makefile .
COPY --link web web
RUN make web/install-deps

# Copy all file
COPY --link . .

# Build frontend
RUN make web/build
```

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# BUILD BACKEND STAGE
+FROM golang:1.18 as backend-builder
+WORKDIR /app
+COPY --from=frontend-builder /app/web/dist /app/web/dist
+COPY . .
```
inductor, line 23:
```suggestion
COPY --link go.mod go.sum ./
RUN go mod download
COPY --link . .
COPY --link --from=frontend-builder /app/web/dist /app/web/dist
```

inductor:
@trungdlp-wolffun added some Dockerfile improvements including latest specification in https://docs.docker.com/engine/reference/builder/#copy---link

Dockerfile
trungdlp-wolffun: Why we should use Node.js 16?

trungdlp-wolffun:
@inductor I'm updated your suggestion. Can you help me understand `COPY --link` meaning ? (I already read Docker document)

inductor:
@trungdlp-wolffun https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#copy---link would probably have a better explanation.

It has better caching than without it

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+FROM golang:1.18 as backend-builder
+WORKDIR /app
+COPY --link go.mod go.sum ./
+RUN go mod download
+COPY --link . .
+COPY --link --from=frontend-builder /app/web/dist /app/web/dist
+RUN make bin
+
+# RUN STAGE
+FROM alpine:latest
```
inductor, line 36: I would avoid using latest tag and might adopt something like distroless base image. But I would like project maintainers to make this decision

trungdlp-wolffun: @inductor I think distroless better than alpine, smaller and only necessary dependency.

trungdlp-wolffun: I updated alpine to ver 3.17

trungdlp-wolffun: @inductor Do you know https://www.chainguard.dev/chainguard-images?

sjbarag:
Any updates here? Having a Dockerfile upstream would make Hermes much easier to deploy!

jeffdaley:
Hey all, thanks for the contributions.

Unfortunately we're not accepting Dockerfile PRs right now, but this is [planned work](https://github.com/hashicorp-forge/hermes/issues?q=is%3Aissue+is%3Aopen+label%3Aplanned-work). Basically, we have internal, build-related things to configure first. 

We appreciate your continued patience on this.

PR #45, opened 2023-02-10
Adds types to FetchService

jeffdaley:
Adds types to the FetchService


web/app/routes/authenticated/drafts.ts
```diff
@@ -93,7 +93,10 @@ export default class DraftsRoute extends Route {
             "/api/v1/drafts?" +
               this.createDraftURLSearchParams(params, ownerFacetOnly)
           )
-          .then((response) => response.json());
+          .then((response) => {
+            assert("response must be defined", response);
```
jfreda, line 97: I'm curious about the uses of `assert` in this PR because assertions are removed for production builds - can you explain the intent behind these?

jeffdaley: To pass type checking and hopefully to pinpoint errors better in development. `resp` can be undefined so we have to assert it or invoke it with a question mark `resp?.json()`. I expect a response so I opted to assert it.

jeffdaley: Thinking about this a little more. I see where you're coming from. It's both more succinct and more true to use the `resp?.json()` invocation. Any benefit of assertion-error proximity is minimal. I'm going to change the approach!

jfreda (approved):

PR #53, opened 2023-02-15
Fix NaN issue in Latest Updates timestamps

jeffdaley:
Fixes an issue where databases with fewer than 4 New/In Review/Published docs could see NaN labels in the Latest Updates tiles.

Long term, we will avoid this by set a `modifiedDate` attribute on doc creation.


jfreda (approved):
Thanks!

PR #55, opened 2023-02-15
Latest Updates cache clearing

jeffdaley:
Adds a `clearCache` function to the AlgoliaService and implements it on the dashboard. Used to override the Algolia JS module's internal cache handling and ensure up-to-date docs in the Latest Updates component.

Current behavior:
Newly In-Review docs don't appear in Latest Updates until the tab is refreshed.

New behavior:
New docs appear in Latest Updates as expected when you load the dashboard or interact with the component.


web/app/components/dashboard/latest-updates.ts
```diff
@@ -59,27 +59,29 @@ export default class DashboardLatestUpdatesComponent extends Component<Dashboard
    * Called onLoad and when tabs are changed.
    */
   fetchDocs = task(async () => {
-    let { currentTab } = this;
```
jeffdaley, line 62: Renamed for clarity

jfreda (approved):
🎉

PR #56, opened 2023-02-16
Scrollable FacetDropdowns with keyboard support

jeffdaley:
- Allows FacetDropdowns to display all filters instead of just the first 10
- Includes a type-to-filter function for long lists, along with a scrollable overflow container
- Improves keyboard-navigation
- Adds documentation and tests


jfreda (approved):
> Adds documentation and tests
😻 

Looks good, and excited for improved ⌨️ support!

PR #57, opened 2023-02-21
Acceptance-test setup with PageTitle tests

jeffdaley:
Configures basic Mirage responses for every route, enough to write PageTitle acceptance tests.


jfreda (approved):
🚀

PR #66, opened 2023-02-28
Fix bug where updating the review status of a document with custom fi…

jfreda:
…elds can error

This bug could cause the indexer to fail in certain cases.

### DIFF
--- a/pkg/models/document_review.go
+++ b/pkg/models/document_review.go
@@ -48,17 +48,9 @@ func (d *DocumentReview) BeforeSave(tx *gorm.DB) error {
 		return err
 	}
 
-	// Get document.
-	if err := d.Document.Get(tx); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/models/document_review.go
```diff
@@ -153,8 +137,30 @@ func (d *DocumentReview) Get(db *gorm.DB) error {
 
 // Update updates the document review in database db.
 func (d *DocumentReview) Update(db *gorm.DB) error {
+	if err := d.getAssociations(db); err != nil {
+		return fmt.Errorf("error getting associations: %w", err)
+	}
+
 	return db.
 		Model(&d).
+		Omit(clause.Associations).
```
jfreda, line 146: Gorm associations magic was the main culprit for the bug.

anubhavmishra (approved):

PR #70, opened 2023-03-06
Use `{{auto-height-textarea}}` modifier on sidebar

jeffdaley:
Adds an `auto-height-textarea` modifier and applies it to the sidebar inputs.

Our current textareas are a hard-coded 4 rows tall. This is overkill for something like "Target version": 

CleanShot 2023-03-06 at 11 19 36@2x

On the other hand, summaries longer than a sentence get cut off:

CleanShot 2023-03-06 at 14 17 50

The new inputs grow (or shrink) according to their content:

CleanShot 2023-03-06 at 14 24 01

CleanShot 2023-03-06 at 14 23 34

with the title and summary fields having a min-height:

CleanShot 2023-03-06 at 14 24 51


jfreda (approved):
🚀

PR #77, opened 2023-03-07
Reset MockDates after tests

jeffdaley:
Adds cleanup code after each use of MockDate.

This isn't affecting us yet because we use the same date in our tests, but this may not always be true. So I'm getting ahead of it.


web/tests/integration/helpers/get-facet-query-hash-test.ts
```diff
@@ -2,8 +2,6 @@ import { module, test } from "qunit";
 import { setupRenderingTest } from "ember-qunit";
 import { render } from "@ember/test-helpers";
 import { hbs } from "ember-cli-htmlbars";
-import MockDate from "mockdate";
-import { equal } from "@ember/object/computed";
```
jeffdaley, line 6: Unused. Found when searching "MockDate"

jfreda (approved):

PR #80, opened 2023-03-08
Add Fetch path to `tsconfig`; update response types

jeffdaley:
Adds the type path for the `ember-fetch` addon.

This revealed a few type errors, which I've fixed here.

Once this is merged #45 resolve quickly.

### DIFF
--- a/web/app/services/authenticated-user.ts
+++ b/web/app/services/authenticated-user.ts
@@ -77,19 +77,19 @@ export default class AuthenticatedUserService extends Service {
    */
   fetchSubscriptions = task(async () => {
     try {
-      let response = await this.fetchSvc.fetch("/api/v1/me/subscriptions", {
-        method: "GET",
-        headers: {
-          "Hermes-Google-Access-Token":
-            this.session.data.authenticated.access_token,
-        },
-      });
-      let subscriptions: string[] = await response.json();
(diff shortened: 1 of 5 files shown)


web/app/services/authenticated-user.ts
```diff
@@ -77,19 +77,19 @@ export default class AuthenticatedUserService extends Service {
...
-      });
-      let subscriptions: string[] = await response.json();
+      let subscriptions = await this.fetchSvc
+        .fetch("/api/v1/me/subscriptions", {
+          method: "GET",
+          headers: {
+            "Hermes-Google-Access-Token":
+              this.session.data.authenticated.access_token,
+          },
+        })
```
jfreda, line 87: Not a blocker, but we shouldn't need this anymore with how we now automatically add the auth header to requests to the backend in the fetch service: https://github.com/hashicorp-forge/hermes/blob/5712506a41213cb11241375c173e53e643334594/web/app/services/fetch.ts#L13-L19

jfreda (approved):

web/app/services/authenticated-user.ts
jeffdaley: Good catch. Addressed separately at https://github.com/hashicorp-forge/hermes/pull/91

PR #82, opened 2023-03-08
Poll for expired session

jeffdaley:
Adds a repeating function to check if the user's auth token has expired.

Shows a flashMessage if it is:

CleanShot 2023-03-08 at 18 52 43@2x


web/app/components/notification.hbs
```diff
@@ -1,14 +1,31 @@
...
+      data-test-flash-notification
+      @flash={{flash}}
+      class="notification"
+      as |component flash close|
+    >
       <Hds::Toast
         @color={{flash.type}}
         @icon={{flash.icon}}
-        @onDismiss={{this.dismiss}}
+        @onDismiss={{close}}
```
jeffdaley, line 12: We used to pass in a no-op based on the now-outdated idea that `dismissOnClick` is always true.

web/app/services/session.ts
```diff
@@ -16,6 +22,79 @@ export function isJSON(str: string) {
...
+  pollForExpiredAuth = keepLatestTask(async () => {
+    await simpleTimeout(Ember.testing ? 100 : 10000);
+
+    this.fetch.fetch(
+      "/api/v1/me",
+      {
+        method: "HEAD",
+        headers: {
+          "Hermes-Google-Access-Token": this.data.authenticated.access_token,
+        },
```
jfreda, line 63: Not strictly necessary because the fetch service adds this automatically for backend requests.

jfreda (approved):
Nice, LGTM

PR #88, opened 2023-03-10
Include `page:1` in filterLink queryParams

jeffdaley:
Includes `page: 1` in filterLink queryParams so that toggling filters effectively creates a new search.

In general, I think this is the expected behavior, but it's worth calling out that it resolves a particularly frustrating use case: Say you're on page 5 of results and you filter by owner. And say the owner has only enough matches for 1 page. Currently by keeping users on page 5, they see a no-data screen and have to manually click page 1. WELL NO MAS, BABY.


jfreda (approved):
Makes sense!

PR #89, opened 2023-03-10
Fix indexer logic by refreshing doc headers before indexing

jfreda:
This PR fixes improves indexer logic by switching the order of operations where we now refresh document(+drafts) headers before indexing the documents for search. Before this, the indexer could get into a state where it was constantly indexing and refreshing the doc headers each run because the header refreshes would cause the documents to have a new modified time.

It's not exactly a hack, but note that we now track the timestamps of when we refresh headers of all documents in a folder by putting a `refreshHeaders:` prefix in the `google_drive_id` column for these records in the `indexer_folders` table.

### DIFF
--- a/internal/indexer/indexer.go
+++ b/internal/indexer/indexer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 	"time"
 
 	validation "github.com/go-ozzo/ozzo-validation/v4"
@@ -64,6 +65,12 @@ type Indexer struct {
 }
 
 type IndexerOption func(*Indexer)
+
+// safeTime is a time.Time that is safe to use concurrently.
+type safeTime struct {
+	time time.Time
+	sync.RWMutex
+}
 
 // NewIndexer creates a new indexer.
 func NewIndexer(opts ...IndexerOption) (*Indexer, error) {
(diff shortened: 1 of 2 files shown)


internal/indexer/indexer.go
```diff
@@ -197,6 +203,126 @@ func (idx *Indexer) Run() error {
...
+			// (if we're indexing it, that is).
+			fd := models.IndexerFolder{
+				GoogleDriveID: fmt.Sprintf("refreshHeaders:%s", idx.DraftsFolderID),
+			}
+			if err := fd.Get(db); err != nil && !errors.Is(
+				err, gorm.ErrRecordNotFound) {
+				log.Error("error getting drafts headers folder indexer data",
+					"error", err,
+				)
+				os.Exit(1)
```
anubhavmishra, line 225: Do we want to exit here? Could we retry instead?

anubhavmishra (changes requested):

anubhavmishra (approved):
Overall looks good, just one minor comment!

internal/indexer/indexer.go
jfreda: It's a good thought, but we're already exiting for basically all errors in the indexer (there's a TODO above to improve error handling). I'd like to address this more holistically in a future PR.

PR #94, opened 2023-03-15
Switch Redirect storage SessionStorage to LocalStorage

jeffdaley:
Moves the user's redirectIntent from SessionStorage to LocalStorage, where it can survive a cross-origin transition. Sets a 2-minute expiration on the cookie to prevent old redirectIntents from sticking around.


jfreda (approved):
We discussed this offline, but just to put it here too - I think this is worth trying to see if it fixes the current issue where redirects aren't preserved in some cases. It's probably not the _best_ way to handle this (the 2 minute expiration sorta-hack), but could be a beneficial interim solution.

jeffdaley:
Working on a PR that prefers sessionStorage but falls back to localStorage.

PR #98, opened 2023-03-21
Authorize contributors to view and patch draft documents

jfreda:
Currently, only doc owners can access a draft document through Hermes even though we share the Google Doc with contributors. This PR allows contributors to view and patch draft documents through Hermes. Delete access is still restricted to document owners.

### DIFF
--- a/internal/api/drafts.go
+++ b/internal/api/drafts.go
@@ -483,11 +483,20 @@ func DraftsDocumentHandler(
 			return
 		}
 
-		// Authorize request (only the owner can access a document draft).
+		// Authorize request (only allow owners or contributors to get past this
+		// point in the handler). We further authorize some methods later that
+		// require owner access only.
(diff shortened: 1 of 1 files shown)


internal/api/drafts.go
```diff
@@ -540,6 +549,14 @@ func DraftsDocumentHandler(
 			l.Info("retrieved document draft", "doc_id", docId)
 
 		case "DELETE":
+			// Authorize request.
```
anubhavmishra, line 552: Interesting, I thought this would already be in place for deleting documents. 😄

anubhavmishra (approved):

internal/api/drafts.go
jfreda: We were! 🙂 We were only authorizing draft owners for all methods before this PR, and now with this PR we authorize owners+contributors for all methods, and this is extra authorization to only allow owners to delete drafts.

PR #102, opened 2023-03-23
Remove deleted drafts from Recently Viewed

jeffdaley:
Adds a `fetchAll` call to the `deleteDraft` task to effectively remove it from the RecentlyViewedDocs list.

Plus some cleanup and boilerplate.


web/app/components/document/index.js
```diff
@@ -1,48 +0,0 @@
...
-          headers: { "Content-Type": "application/json" },
-        })
-        .then((resp) => resp.json())
-        .catch((err) => {
-          console.log(`Error deleting document draft: ${err}`);
-          throw err;
-        });
-
-      // Wait for document to be deleted.
-      yield wait(2000);
```
jeffdaley, line 29: I don't think we need to wait for this since we're redirecting to a known route (unlike the [new doc form](https://github.com/hashicorp-forge/hermes/blob/main/web/app/components/new/doc-form.ts#L219)).

web/app/components/document/index.js
```diff
@@ -1,48 +0,0 @@
...
-    // Returns a promise that always results after the provided number of milliseconds
-    const wait = (ms) => new Promise((res) => setTimeout(res, ms));
-
-    try {
-      const docResp = yield this.fetchSvc
-        .fetch("/api/v1/drafts/" + docID, {
-          method: "DELETE",
-          headers: { "Content-Type": "application/json" },
-        })
-        .then((resp) => resp.json())
```
jeffdaley, line 22: Doesn't look like we're doing anything with the JSON. Removed in TS version

web/mirage/config.ts
```diff
@@ -146,8 +146,6 @@ export default function (mirageConfig) {
        * Used by the Document route to get a document.
        */
       this.get("/documents/:document_id", (schema, request) => {
-        debugger;
```
jeffdaley, line 149: 😅

web/app/services/recently-viewed-docs.ts
```diff
@@ -308,3 +308,9 @@ export default class RecentlyViewedDocsService extends Service {
     }
   );
 }
+
+declare module "@ember/service" {
+  interface Registry {
+    "recently-viewed-docs": RecentlyViewedDocsService;
+  }
```
jeffdaley, line 315: This is to allow aliasing in TS files, e.g., `@service('recently-viewed-docs') declare recentDocs;`

jfreda (approved):

PR #103, opened 2023-03-23
Combine Session and LocalStorage for redirect handling

jeffdaley:
A follow up to #94 (and response to @jfreda's [comment](https://github.com/hashicorp-forge/hermes/pull/94#pullrequestreview-1355716661)),
this adds localStorage as a fallback redirect solution rather the primary one. 

Still need to figure out why SessionStorage doesn't work on the first try...


web/app/services/session.ts
```diff
@@ -13,17 +13,20 @@ export default class SessionService extends EmberSimpleAuthSessionService {
...
 
     let redirectTarget: string | null = null;
     let transition;
 
-    if (redirectObject) {
-      // Check if the object is less than 2 minutes old
-      if (Date.now() < JSON.parse(redirectObject).expiresOn) {
-        redirectTarget = JSON.parse(redirectObject).url;
+    if (redirectStorageValue) {
+      const isObject = redirectStorageValue.startsWith("{");
```
jfreda, line 24: Is this an idiomatic way to test for this, or would it maybe be better to use `typeof` or something?

jfreda (approved):

web/app/services/session.ts
jeffdaley: `typeof` was also my first thought, but all localStorage/sessionStorage values are saved as strings.

I've updated to use `JSON.parse` instead of the frail `startWith` method.

jfreda: Got it, makes sense!

PR #118, opened 2023-04-05
WIP: Increase facet limit; Add loader for long lists

jeffdaley:
Increases the Algolia maxFacets value to 1000 — [the highest number](https://www.youtube.com/watch?v=9P2ROAbQZYw&themeRefresh=1) — so that more document owners are searchable via the filter dropdowns.

And because long lists can take a second to load, I've added and tested a loading function for lists longer than 250. This is a short-term fix while we explore more robust solutions that can handle more than 1000 authors and display them quicker.

Also fixes a truncation bug on longer owner names.

### DIFF
--- a/web/app/components/header/facet-dropdown.hbs
+++ b/web/app/components/header/facet-dropdown.hbs
@@ -30,24 +30,34 @@
     }}
   />
   {{#if this.dropdownIsShown}}
-    <Header::FacetDropdownList
-      {{dismissible dismiss=this.hideDropdown related=this.triggerElement}}
-      @shownFacets={{this.shownFacets}}
-      @label={{@label}}
-      @inputIsShown={{this.inputIsShown}}
-      @onInput={{perform this.onInput}}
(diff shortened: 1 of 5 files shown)


web/app/components/header/facet-dropdown.hbs
```diff
@@ -30,24 +30,34 @@
...
-      @currentSortByValue={{@currentSortByValue}}
-    />
+    {{#if this.loaderIsShown}}
+      <div
+        class="facet-dropdown-popover px-4 pt-3 pb-3.5 large hds-dropdown-list text-color-foreground-faint text-body-200"
+        data-test-facet-dropdown-loader
+        {{did-insert (perform this.removeLoader)}}
+      >
+        Loading...
+      </div>
```
jeffdaley, line 40: As soon as this is inserted, we remove it. However, because the FacetDropdownList has a long render time (~1000ms), the loader is not visually removed until the dropdown list can replace it.

jeffdaley:
Closing this while we consider other solutions.

PR #125, opened 2023-04-06
Add shortLinkBaseURL checks (trailingSlack, isValidURL) and test them

jeffdaley:
Adds some safeguarding to the ShortLinkBaseURL functions: We now have trailing-slash logic and we confirm the URL is valid before using it.


jfreda (approved):
Nice, test cases look good!

PR #131, opened 2023-04-10
Extend Torii `restore` method; Remove unnecessary `invalidate` calls

jeffdaley:


web/app/routes/application.js
```diff
@@ -17,7 +17,6 @@ export default class ApplicationRoute extends Route {
   error(error) {
     if (error instanceof UnauthorizedError) {
       this.session.invalidate();
-      this.router.transitionTo("authenticate");
```
jeffdaley, line 20: Happens automatically via invalidate

web/app/routes/authenticated.ts
```diff
@@ -7,7 +7,7 @@ export default class AuthenticatedRoute extends Route {
   @service declare session: SessionService;
   @service declare authenticatedUser: AuthenticatedUserService;
 
-  async beforeModel(transition: any) {
```
jeffdaley, line 10: No longer async

web/app/services/fetch.ts
```diff
@@ -33,13 +42,9 @@ export default class FetchService extends Service {
       }
 
       if (!resp.ok) {
-        if (resp.status === 401) {
-          if (isPollCall) {
-            // handle poll-call failures via the session service
-            return;
-          } else {
-            this.session.invalidate();
```
jeffdaley, line 41: Since we invalidate on applicationError, calling it here seems redundant at best.

web/app/services/session.ts
```diff
@@ -112,7 +112,7 @@ export default class SessionService extends EmberSimpleAuthSessionService {
       }
     }
 
-    if (redirectTarget) {
+    if (redirectTarget && redirectTarget !== "/authenticate") {
```
jeffdaley, line 115: `redirectTarget` becomes `authenticate` when refreshing the login screen. In this case, we ignore it and fall back to the `routeAfterAuthentication` value of `'index'`.

web/app/authenticators/torii.ts
```diff
@@ -0,0 +1,28 @@
...
+  @service declare torii: unknown;
+
+  async restore(_arguments: unknown) {
+    const data = await super.restore(_arguments);
+    /**
+     * Try the restored credentials with the backend.
+     * If the backend rejects the credentials, the error will bubble up
+     * to the application route's error method, which invalidates the session.
+     */
+    await this.fetchSvc.fetch("/api/v1/me", {
```
meirish, line 19: This is kinda nuanced but I'm not sure if you specifically need to return the promise here - I thought that's how the restore / authenticate hooks worked. If this throws, but you're not returning the rejected promise, I don't know if it will behave properly.

meirish (commented):
Had one question about authenticator hooks, but otherwise makes sense!

web/app/authenticators/torii.ts
jfreda: @meirish Do you mean that an error here won't bubble up to the application route? (which I _think_ is how this is intended to be handled if I'm understanding the logic correctly)

web/app/routes/application.js
jfreda: I've lost context for this... where is that route configured then for `invalidate`? 🤔

web/app/authenticators/torii.ts
jeffdaley: @jfreda You have the right idea. My thought is that this failed call will return a 401 which we let bubble up to application. 

@meirish Are you suggesting that if `restore` doesn't return a promise, we might see unintended consequences?

--

Not to make this more confusing, but, it looks like `restore` has its own rejection handling. I don't think I was right about it bubbling to the error route.

> A rejecting promise indicates that data does not constitute a valid session and will result in the session being invalidated or remaining unauthenticated.

https://ember-simple-auth.com/api/classes/BaseAuthenticator.html#method_restore

web/app/routes/application.js
jeffdaley: Not 100%, but I think it automatically redirects to `/` which in our app is `/authenticated`, which has a `requireAuthentication` check that, if failed, redirects to `authenticate`.

web/app/authenticators/torii.ts
meirish: Yeah the method should return a promise, and on rejection, it kicks you to the route configured for authentication. I’m not sure of the behavior when returning an object or the api call erroring - which are the two cases outlined here.

jfreda (approved):
Looks reasonable to me - let's give it a try! 🤞

web/app/authenticators/torii.ts
```diff
@@ -15,11 +15,17 @@ export default class ToriiAuthenticator extends Torii {
      * A rejecting promise indicates invalid session data and will result
      * in the session being invalidated or remaining unauthenticated.
      */
-    return this.fetchSvc.fetch("/api/v1/me", {
+    await this.fetchSvc.fetch("/api/v1/me", {
```
meirish, line 18: So this needs to return a promise, and that promise, upon success, needs to return the data that will be set on the session service.  You can combine the two calls so it has the expected return value, and resolve value with something like this:

```
const data = await super.restore(...arguments);
    /**
     * A rejecting promise indicates invalid session data and will result
     * in the session being invalidated or remaining unauthenticated.
     */
    return this.fetchSvc.fetch("/api/v1/me", {      method: "HEAD",
      headers: {
        "Hermes-Google-Access-Token": data.access_token,
      },
    }).then(() => return data);

```

jeffdaley: Good call!

PR #132, opened 2023-04-12
Add API endpoint for recently viewed docs

jfreda:
This PR adds an API endpoint `/api/v1/me/recently-viewed-docs` to retrieve recently viewed docs for a logged-in user. Now the source of truth for recently viewed docs is the database, whereas Hermes previously got this information from a JSON file stored in Google Drive's application data - this PR doesn't remove updating the data there yet (this will be done shortly in another PR).

Additionally a bug in the logic for updating the recently viewed docs in the database was fixed: the recently viewed documents were being updated correctly, but there was the potential for the wrong document's "viewed at" time to be modified. This resulted the dashboard's recently viewed documents section to sometimes be displayed out of order.

### DIFF
--- a/internal/api/documents.go
+++ b/internal/api/documents.go
@@ -159,64 +159,15 @@ func DocumentHandler(
 				// Get authenticated user's email address.
 				email := r.Context().Value("userEmail").(string)
 
-				// Get user (if exists).
-				u := models.User{
-					EmailAddress: email,
-				}
-				if err := u.Get(db); err != nil && !errors.Is(
-					err, gorm.ErrRecordNotFound) {
-					l.Error("error getting user in database",
+				if err := updateRecentlyViewedDocs(email, docID, db, now); err != nil {
+					// If we get an error, log it but don't return an error response
+					// because this would degrade UX.
+					l.Error("error updating recently viewed docs",
 						"error", err,
 						"doc_id", docID,
 						"method", r.Method,
 						"path", r.URL.Path,
 					)
-					// TODO: return an error response when this is required.
-					// http.Error(w, "Error requesting document",
-					// 	http.StatusInternalServerError)
-					return
-				}
-
-				// Prepend document to recently viewed documents.
-				rvd := append(
-					[]models.Document{{GoogleFileID: docID}},
-					u.RecentlyViewedDocs...)
-
-				// Trim recently viewed documents to a length of 5.
-				if len(rvd) > 5 {
-					rvd = rvd[:5]
-				}
-
-				// Update user.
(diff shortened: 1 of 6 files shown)


web/app/services/recently-viewed-docs.ts
```diff
@@ -74,18 +74,7 @@ export default class RecentlyViewedDocsService extends Service {
...
-            fields: "files(id, name)",
-          }),
-        {
-          headers: {
-            Authorization:
-              "Bearer " + this.session.data.authenticated.access_token,
-            "Content-Type": "application/json",
-          },
-        }
+        "/api/v1/me/recently-viewed-docs"
```
jeffdaley, line 77: 😍

internal/api/me_recently_viewed_docs.go
```diff
@@ -0,0 +1,129 @@
...
+		switch r.Method {
+		case "GET":
+			// Find or create user.
+			u := models.User{
+				EmailAddress: userEmail,
+			}
+			if err := u.FirstOrCreate(db); err != nil {
+				errResp(
+					http.StatusInternalServerError,
+					"Error authorizing the request",
```
anubhavmishra, line 57: nit: maybe "Error validating user information in request" would be more explicit of what error we want to show the user?

internal/api/documents.go
```diff
@@ -394,3 +345,55 @@ Hermes
...
+		return fmt.Errorf("error getting user in database: %w", err)
+	}
+
+	// Prepend document to recently viewed documents.
+	rvd := append(
+		[]models.Document{{GoogleFileID: docID}},
+		u.RecentlyViewedDocs...)
+
+	// Trim recently viewed documents to a length of 5.
+	if len(rvd) > 5 {
```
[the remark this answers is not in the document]
jfreda: The frontend does trim this as well. This is to limit the amount of recently viewed docs that we store in the database.

internal/api/me_recently_viewed_docs.go
jfreda: This is more of an internal error where we've got an issue with the user data in our database vs. any issues with user info in the request (which is validated by the `auth` package), which is why I kind of favor using an internal error and this message instead of a 403.

anubhavmishra: Sounds reasonable!

internal/api/documents.go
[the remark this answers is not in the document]
anubhavmishra: I see, makes total sense now.

anubhavmishra (approved):

PR #134, opened 2023-04-12
Auth: Ignore saving certain transitions

jeffdaley:
Ignores post-auth redirectTargets to the root and Authenticate routes. Fixes a bug causing valid redirects to be overwritten.

When you try to access an authenticated route, Ember Simple Auth restores your previous token, which, if expired, will cause the session to invalidate. By default, this [reloads the Ember app](https://ember-simple-auth.com/api/classes/SessionService.html#method_handleInvalidation) in an effort to clear data. On refresh, the application beforeModel receives a new transition, this time to the application root. Currently we let this transition target through, overwriting the valid target. Now we ignore it :----)

### DIFF
--- a/web/app/services/session.ts
+++ b/web/app/services/session.ts
@@ -112,7 +112,7 @@ export default class SessionService extends EmberSimpleAuthSessionService {
       }
     }
 
-    if (redirectTarget && redirectTarget !== "/authenticate") {
+    if (redirectTarget) {
       transition = this.router.transitionTo(redirectTarget);
     } else {
       transition = this.router.transitionTo(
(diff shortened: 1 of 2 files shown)


web/app/services/session.ts
```diff
@@ -112,7 +112,7 @@ export default class SessionService extends EmberSimpleAuthSessionService {
       }
     }
 
-    if (redirectTarget && redirectTarget !== "/authenticate") {
```
jeffdaley, line 115: Moved this to the beforeModel check

jfreda (approved):

PR #141, opened 2023-04-14
Improve `pollForExpiredAuth` and reauthenticate UX

jeffdaley:
Improves `pollForExpiredAuth` handling as well as the reauthentication UX:

- We now refresh userInfo as part of the re-auth flow
- We now show success and error messages on re-auth
- We now check for stale reauthMessages during polling. Now, e.g.,  if someone reauthenticates on another tab, the other tabs should update within 10 seconds.

Plus some tests and cleanup.


web/tests/acceptance/application-test.ts
```diff
@@ -84,23 +88,41 @@ module("Acceptance | application", function (hooks) {
...
-  test("the authorize button works as expected", async function (this: ApplicationTestContext, assert) {
+  test("the reauthenticate button works as expected (success)", async function (this: ApplicationTestContext, assert) {
     let authCount = 0;
 
     await authenticateSession({});
 
-    this.session.authenticate = () => {
+    this.session.authenticate = async () => {
       authCount++;
     };
```
jeffdaley, line 98: This isn't async, but we do it to keep type compliance with the service.

web/app/services/session.ts
```diff
@@ -55,16 +66,18 @@ export default class SessionService extends EmberSimpleAuthSessionService {
...
     await simpleTimeout(Ember.testing ? 100 : 10000);
 
-    this.fetch.fetch(
-      "/api/v1/me",
-      {
-        method: "HEAD",
-      },
-      true
-    );
+    // If the reauth message is shown, do nothing but restart the counter.
```
jfreda, line 69: I'm thinking of a situation where a user has multiple Hermes tabs open and the session expires: can/should we continue polling to check to see if the `me` API starts returning a 200 again? This way when the user re-auths in one tab, the flash message will go away in all the other tabs?

jfreda (approved):
Looks good! Thanks for addressing the Okta reauth use case!

PR #143, opened 2023-04-17
Reconfigure Ember Metrics / Google Analytics

jeffdaley:
Updates `ember-metrics` to allow configuration via the ConfigService as an alternative to environment variables.

### DIFF
--- a/web/app/metrics-adapters/google-analytics-four.ts
+++ b/web/app/metrics-adapters/google-analytics-four.ts
@@ -1,7 +1,6 @@
 import { inject as service } from "@ember/service";
 import ConfigService from "hermes/services/config";
 import GoogleAnalyticsFour from "ember-metrics/metrics-adapters/google-analytics-four";
-import { compact } from "ember-metrics/-private/utils/object-transforms";
 
 declare global {
   interface Window {
@@ -36,6 +35,6 @@ export default class GoogleAnalyticsFourAdapter extends GoogleAnalyticsFour {
     window.dataLayer = window.dataLayer || [];
 
     this.gtag("js", new Date());
-    this.gtag("config", GoogleAnalyticsTagID, compact(this.options));
+    this.gtag("config", GoogleAnalyticsTagID, this.options);
   }
 }
--- a/web/app/metrics-adapters/google-analytics-four.ts
+++ b/web/app/metrics-adapters/google-analytics-four.ts
@@ -17,7 +17,7 @@ export default class GoogleAnalyticsFourAdapter extends GoogleAnalyticsFour {
 
   /**
    * Modified from its parent class:
-   * https://github.com/adopted-ember-addons/ember-metrics/blob/master/addon/metrics-adapters/google-analytics-four.js#L24-L28
+   * https://github.com/adopted-ember-addons/ember-metrics/blob/master/addon/metrics-adapters/google-analytics-four.js#L11-L28
    */
   install(): void {
     const GoogleAnalyticsTagID = this.configSvc.config.google_analytics_tag_id;
--- a/web/app/metrics-adapters/google-analytics-four.ts
+++ b/web/app/metrics-adapters/google-analytics-four.ts
@@ -13,7 +13,7 @@ export default class GoogleAnalyticsFourAdapter extends GoogleAnalyticsFour {
 
   /**
    * Modified from its parent class:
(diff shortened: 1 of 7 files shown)


web/config/environment.js
```diff
@@ -34,20 +32,15 @@ module.exports = function (environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
-    metricsAdapters: hasGoogleAnalyticsID
-      ? [
-          {
-            name: "GoogleAnalyticsFour",
-            environments: ["development", "production"],
-            config: {
-              id: getEnv("GOOGLE_ANALYTICS_TAG_ID"),
```
anubhavmishra, line 43: Curious if setting tag ID via the `GOOGLE_ANALYTICS_TAG_ID` environment variable will be still supported?

web/app/services/config.ts
```diff
@@ -12,6 +12,7 @@ export default class ConfigService extends Service {
     feature_flags: config.featureFlags,
     google_doc_folders: config.google.docFolders ?? "",
     short_link_base_url: config.shortLinkBaseURL,
+    google_analytics_tag_id: config.googleAnalyticsTagId
```
jfreda, line 15: Hmm, does this work? It looks like `config.googleAnalyticsTagId` was removed in this PR [here](https://github.com/hashicorp-forge/hermes/pull/143/files#diff-5d61833860ee05d29b82a217c3ad8a57ac5ce86bf8e07e486b94bad91b37ec8cL14)?

web/app/metrics-adapters/google-analytics-four.ts
```diff
@@ -0,0 +1,39 @@
...
+    dataLayer: any[];
+  }
+}
+
+export default class GoogleAnalyticsFourAdapter extends GoogleAnalyticsFour {
+  @service("config") declare configSvc: ConfigService;
+
+  /**
+   * Modified from its parent class:
+   * https://github.com/adopted-ember-addons/ember-metrics/blob/master/addon/metrics-adapters/google-analytics-four.js#L11-L28
```
jfreda, line 16: Nit: might want to put the commit ID here in case future changes to this file in this branch change the line numbers.

```suggestion
   * https://github.com/adopted-ember-addons/ember-metrics/blob/01bb65e68f565fc48c2fa252d17a1b7d3d099e68/addon/metrics-adapters/google-analytics-four.js#L11-L28
```

jfreda (commented):
One question, but overall looking good!

web/app/metrics-adapters/google-analytics-four.ts
```diff
@@ -11,23 +11,21 @@ declare global {
...
+    const GoogleAnalyticsTagID =
+      this.config.id || this.configSvc.config.google_analytics_tag_id;
 
     if (!GoogleAnalyticsTagID) {
       return;
     }
 
     this.options = {
       id: GoogleAnalyticsTagID,
+      send_page_view: true,
```
anubhavmishra, line 28: 🎉

web/app/routes/application.ts
```diff
@@ -77,13 +68,11 @@ export default class ApplicationRoute extends Route {
...
 
-    // Flags read from the environment and set properties on the service this
-    // could be done in an initializer, but this seems more natural these days
     this.flags.initialize();
 
-    // Get web config from backend if this is a production build.
+    // Set config from the backend in production
     if (config.environment === "production") {
-      return this.fetchSvc
+      await this.fetchSvc
```
anubhavmishra, line 75: Good call!

anubhavmishra (approved):

anubhavmishra:
Overall, looks good to me too!

web/app/services/config.ts
jeffdaley: That line was just the check for whether the variable existed, from when the config was simpler.

Now [we pass in that variable regardless of whether it exists](https://github.com/hashicorp-forge/hermes/pull/143/files#diff-5d61833860ee05d29b82a217c3ad8a57ac5ce86bf8e07e486b94bad91b37ec8cR40). Previously, that would throw errors; but now we have code to [deal with that scenario](https://github.com/hashicorp-forge/hermes/pull/143/files#diff-0eb08ae4ac297fb24e5c9a69be04f4197442d01e06d7b822b325bcc673923fd3R19-R24).

jfreda: I wonder if we wouldn't need to override that class if this line was something like:
```
google_analytics_tag_id: config.metricsAdapters[0].config.id
```
(there's probably a way to make this less brittle than just referencing the first element too?)

PR #144, opened 2023-04-18
Bypass Google auth when Okta is configured

jfreda:
This PR adds functionality to bypass Google auth when Okta is configured. To do this, a GET method was added for the  `/api/v1/me` API with a response that mimics Google's `userinfo` API, which we previously used to get a user's profile photo and name. Also, now that #132 was merged, this removes the code for using the Google Drive application data for storing recently viewed docs.

## Breaking changes

- The Admin SDK API now needs to be enabled.
- New required [aws_region](https://github.com/hashicorp-forge/hermes/pull/144/commits/e6bd4ade8ee16d79507368a4a64ed1ab317f9720#diff-ab2c1b25e9bbe647fba0008a91e18a43a3130568bafddd04f33fb619341a4eaeR135-R136) configuration option for the `okta` block


web/web.go
```diff
@@ -57,6 +57,7 @@ type ConfigResponse struct {
 	AlgoliaDocsIndexName     string          `json:"algolia_docs_index_name"`
 	AlgoliaDraftsIndexName   string          `json:"algolia_drafts_index_name"`
 	AlgoliaInternalIndexName string          `json:"algolia_internal_index_name"`
+	BypassGoogleAuth         bool            `json:"bypass_google_auth"`
```
anubhavmishra, line 60: This is a nit, but I wonder if we could use "skip" instead of "bypass"? For example, `SkipGoogleAuth`, `skip_google_auth`, `SKIP_GOOGLE_AUTH `?

web/web.go
```diff
@@ -99,10 +100,16 @@ func ConfigHandler(
 			shortLinkBaseURL = strings.TrimSuffix(cfg.BaseURL, "/") + "/l"
 		}
 
+		bypassGoogleAuth := false
```
anubhavmishra, line 103: Maybe an explicit description of when the Google Auth would be skipped in a comment would help. Also, adding a note in the [configs/config.hcl](https://github.com/hashicorp-forge/hermes/blob/main/configs/config.hcl) file where Okta config is specified would help other users as well.

web/web.go
jfreda: Done! (I didn't have any preference on this wording)

web/web.go
jfreda: Done!

anubhavmishra: I don't see a note in the [configs/config.hcl](https://github.com/hashicorp-forge/hermes/blob/main/configs/config.hcl) file where Okta config is specified about Google auth being skipped.

anubhavmishra: Ah, maybe I wasn't clear in my comments sorry about that. I was suggesting that we leave a note in form of a comment in the [configs/config.hcl](https://github.com/hashicorp-forge/hermes/blob/main/configs/config.hcl) file near the [Okta config block](https://github.com/hashicorp-forge/hermes/blob/main/configs/config.hcl#L131) to describe that when Okta is configured then Google Auth would be skipped. Does that make sense?

jfreda: Got it, updated https://github.com/hashicorp-forge/hermes/pull/144/commits/be6eb3f2973edeec9e0aa62d5db79ecc4264ca29

web/app/routes/authenticated/document.js
```diff
@@ -106,10 +105,13 @@ export default class DocumentRoute extends Route {
...
-
-    // Record the doc with the RecentlyViewedDocs service.
-    void this.recentDocs.markViewed.perform(params.document_id, params.draft);
+    // Update recently viewed docs for the dashboard route.
+    try {
+      await this.recentDocs.fetchAll.perform();
+    } catch (e) {
+      console.error("error updating recently viewed docs", e);
+      throw e;
+    }
```
jeffdaley, line 114: This doesn't seem important here. What if you remove it and defer that load for when the user is actually visiting the dashboard? Otherwise I'd call it with `void` instead of `try { await }` so it doesn't hold up loading other stuff.

web/app/components/header/nav.hbs
```diff
@@ -81,11 +81,13 @@
...
-            {{on "click" this.invalidateSession}}
-            @text="Sign out"
-          />
+          {{#if this.showSignOut}}
+            <dd.Interactive
+              data-test-user-menu-item="sign-out"
+              {{on "click" this.invalidateSession}}
+              @text="Sign out"
+            />
+          {{/if}}
```
jeffdaley, line 90: Why are we removing the logout button? What happens if I `skipGoogleAuth` and try to `invalidateSession`?

web/app/routes/authenticate.ts
```diff
@@ -1,15 +1,19 @@
...
 
   beforeModel() {
     /**
-     * Checks if the session is authenticated,
+     * If using Google auth, checks if the session is authenticated,
      * and if it is, transitions to the specified route
      */
-    this.session.prohibitAuthentication("/");
+    if (!this.configSvc.config.skip_google_auth) {
+      this.session.prohibitAuthentication("/");
```
jeffdaley, line 16: What happens if remove this `if` statement and run `prohibitAuthentication` under `skipGoogleAuth` conditions?

web/app/routes/authenticated.ts
```diff
@@ -26,9 +30,10 @@ export default class AuthenticatedRoute extends Route {
     await this.authenticatedUser.loadInfo.perform();
 
     /**
-     * If the session is authenticated with the front- and back-ends,
-     * kick off the task to poll for expired auth.
+     * If using Google auth, kick off the task to poll for expired auth.
      */
-    void this.session.pollForExpiredAuth.perform();
+    if (!this.configSvc.config.skip_google_auth) {
+      void this.session.pollForExpiredAuth.perform();
```
jeffdaley, line 36: Why does skipping Google auth negate the need to poll for expiredAuth?

jeffdaley (commented):
Looks good. I have some questions but it seems like we're close.

We seem light on front-end tests given the number of changes. Depending on your answers, let's stub some todos if not get a few more tests in here. I can help with that of course.

web/app/components/header/nav.hbs
jfreda: It doesn't do anything because there's no Ember Simple Auth session to invalidate when we bypass Google auth.

web/app/routes/authenticated/document.js
jfreda: I was just minimizing changes, and when I fully removed this, the recently viewed docs weren't being updated in the dashboard until I refreshed the page. I just tried moving this to the dashboard route though and it seems to work, so I can move it there.

web/app/routes/authenticate.ts
jfreda: Looks like it's the same behavior. I think I forgot what this method actually does - I can revert this change.

web/app/routes/authenticated.ts
jfreda: Good q, 2 things here:
- I still need to do some more testing on the app behavior when Okta auth expires; it might still be useful to do this polling if the user isn't immediately redirected to log in to Okta.
- Currently `pollForExpiredAuth` is (understandably) Google-centric so I think we'll want to at least skip this until that either becomes neutral to the auth provider or has an Okta option built in too (assuming that we'll want to poll when skipping Google auth; still need to test).

web/app/components/header/nav.hbs
jeffdaley: Can we have it invalidate with Okta instead? I assume users will still want to sign out.

web/app/routes/authenticate.ts
jeffdaley: It's supposed to redirect you to `/` if you're authenticated with Ember Simple Auth

jfreda: Yeah, actually I guess we should probably just redirect right away from this route if we're bypassing Google auth... Because users could still have this URL saved and it won't be relevant without Google auth.

web/app/routes/authenticated.ts
jeffdaley: > it might still be useful to do this polling if the user isn't immediately redirected to log in to Okta

Yeah, I'd like to maintain our existing behavior where we detect passive logouts: If your token expires in the background, we should block [Okta]'s immediate-redirect function to show a gentler "reauathenticate" message.

Your change makes sense and its good to know you're planning to test further.

web/app/components/header/nav.hbs
jfreda: Our supported Okta configuration enforces Okta auth at the load balancer level so there's not going to be a way to sign out unless the user signs out of their Okta session (which will be in Okta, not Hermes).

jeffdaley: Can we change our supported configuration to allow logout? 😬 I'd really like to keep the log out button. "As a user I want to log out."

web/app/routes/authenticate.ts
jfreda: https://github.com/hashicorp-forge/hermes/pull/144/commits/8d11b85038a6155bb28efc9c6ea4c6853272a247 implements this ^

web/app/routes/authenticated/document.js
jfreda: https://github.com/hashicorp-forge/hermes/pull/144/commits/8143ed4ec9ae99a5439578ce3cf49ec30c6020fb

web/app/routes/authenticated.ts
jfreda: Re-enabled polling auth for Okta in this commit: https://github.com/hashicorp-forge/hermes/pull/144/commits/23635e630b6d1407d67b129b907fa7d8dc1273b6

web/app/components/header/nav.hbs
jfreda: We chatted about this offline, but I think I found a way to do this and it'll be implemented in another PR soon!

PR #155, opened 2023-04-26
Try a /document redirect on /draft fetchError

jeffdaley:
Adds a retry-draft-as-published-document function to the document model.

Currently, once a draft is published, you can no longer reach it by its former URL (the one you probably have in your browser history). The `?draft=true` param throws off the model hook and you get redirected to the dashboard.

Now, we catch draft-fetch failures and retry them as published-doc fetches.


jfreda (approved):
This is a nice UX improvement 👍

PR #156, opened 2023-04-26
Clean up `Doc` components

jeffdaley:
Cleans the Doc components of now-redundant methods. Simplifies the CSS a bit.


web/app/components/doc/tile.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default class DocTileComponent extends Component<DocTileComponentSignature> {
+  protected get productAreaName(): string | undefined {
+    switch (this.args.productArea) {
+      case "Cloud Platform":
+        return "HCP";
+      default:
+        return this.args.productArea;
+    }
```
jfreda, line 27: This is definitely an improvement over the previous code, but longer-term we should think how we can make this more generic (might require some data provided by the backend) 🤔

jfreda (approved):

web/app/components/doc/tile.ts
jeffdaley: Yeah. Maybe an attribute like `flightIconName` or `shortName`.

PR #166, opened 2023-05-02
Refactor FacetDropdownList

jeffdaley:
Converts the FacetDropdownList components to use `X::DropdownList`.


web/app/components/header/toolbar.hbs
```diff
@@ -14,12 +14,13 @@
               @label="Type"
               @facets={{@facets.docType}}
               @disabled={{not @facets.docType}}
+              class=""
```
jfreda, line 17: Any reason for this empty class attribute? (more curious vs. strong feelings against it)

jfreda (approved):
So much red! 😻

PR #168, opened 2023-05-03
Use http.StatusOK for w.WriteHeader

syntaqx:
Minor little fix, but keeps the `w.WriteHeader` the same as other implementations in the code, and also because using `200` without the constant value makes me sad.


jfreda (approved):
Thanks!

> using 200 without the constant value makes me sad.

PR #170, opened 2023-05-04
Configure Torii via the web config API

jfreda:
[Torii](https://github.com/FMGSuite/torii) only gets its configuration from the Ember environment config, which required these values to be set and then baked into the web app (which was a blocker for us being able to publish binaries for Hermes). This overrides the appropriate Torii initializer so it fetches the Google OAuth configuration from the web config API (for production builds).


jeffdaley (approved):
Aces

PR #171, opened 2023-05-05
Only disable requesting a review via checkbox value when a check is configured

jfreda:


web/app/components/document/sidebar.hbs
```diff
@@ -517,7 +514,10 @@
       @taskButtonText="Request review"
       @taskButtonLoadingText="Requesting review..."
       @taskButtonIcon="test"
-      @taskButtonIsDisabled={{not this.docTypeCheckboxValue}}
+      @taskButtonIsDisabled={{and
+        (not this.docTypeCheckboxValue)
+        @docType.checks.length
+      }}
```
jfreda, line 520: This is the actual fix

PR #175, opened 2023-05-16
Add support for FloatingUI offset

jeffdaley:
Adds support for custom FloatingUI offsets. Will be used by the search popover to align icons and type.

https://floating-ui.com/docs/offset


jfreda (approved):
Nice, LGTM

PR #176, opened 2023-05-16
Add `maybe-query` and `model-or-models` helpers

jeffdaley:
Introduces two LinkTo-focused helpers to make the DropdownList::LinkTo component more flexible.

Example usage:
```
<LinkTo
    {{! fall back to an empty hash if @query is undefined }}
    @query={{maybe-query @query}}
    
    {{! fall back to an empty array if no models are provided }}
    @models={{model-or-models @model @models}}
```

### DIFF
--- a/web/app/helpers/maybe-query.ts
+++ b/web/app/helpers/maybe-query.ts
@@ -1,11 +1,28 @@
 import { helper } from "@ember/component/helper";
 
+export interface MaybeQuerySignature {
+  Args: {
+    Positional: [Record<string, unknown> | undefined];
+  };
+  Return: Record<string, unknown> | {};
+}
+
 /**
  * Supplies an empty object if no is query provided.
  * Avoids errors when passing empty `@query` values to LinkTos.
  * Workaround for https://github.com/emberjs/ember.js/issues/19693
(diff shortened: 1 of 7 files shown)


web/app/helpers/maybe-query.ts
```diff
@@ -0,0 +1,28 @@
...
+    Positional: [Record<string, unknown> | undefined];
+  };
+  Return: Record<string, unknown> | {};
+}
+
+/**
+ * Supplies an empty object if no is query provided.
+ * Avoids errors when passing empty `@query` values to LinkTos.
+ * Workaround for https://github.com/emberjs/ember.js/issues/19693
+ * Can be removed when we upgrade to Ember 3.28+
```
jfreda, line 14: Hmm looks like we're already on Ember 3.28? https://github.com/hashicorp-forge/hermes/blob/86f624b3c8212a59fc5e3af36e6680049f169206/web/package.json#L67

From that linked issue, I think we might have to upgrade to v4 for this to be fixed? (if that's correct, I'm suggesting that we just update the comment, and not that we perform the upgrade for this PR 🙂)

jfreda (approved):

PR #178, opened 2023-05-17
WIP: Show a message when a document is in a "bad state"

jeffdaley:
CleanShot 2023-05-17 at 12 01 09@2x

TODO: Determine criteria for this.

### DIFF
--- a/web/app/components/document/sidebar.hbs
+++ b/web/app/components/document/sidebar.hbs
@@ -351,47 +351,48 @@
     </div> --}}
 
     {{#if this.userHasEditPrivileges}}
-      <div class="sidebar-footer">
-        {{#if (not @document.appCreated)}}
+      <div
+        class="sidebar-footer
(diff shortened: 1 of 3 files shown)


web/app/components/document/sidebar.hbs
```diff
@@ -351,26 +351,48 @@
...
               </div>
-              <div class="text-[12px] opacity-60 italic">Create docs in-app for
-                best results</div>
+              {{#if this.docStateIsBad}}
+                <Hds::Link::Standalone
+                  @icon="arrow-right"
+                  @iconPosition="trailing"
+                  @text="See issues on GitHub"
+                  @size="small"
+                  @href="https://github.com/hashicorp-forge/hermes/issues"
```
anubhavmishra, line 389: Let's think more about this one. I would like us to point to the exact bug and a possible solution for the users.

Issue #181, opened 2023-05-25
Google Docs DeleteContentRange API fails when content contains a suggestion


jfreda:
There is currently a bug with the Google Docs [DeleteContentRange API](https://developers.google.com/docs/api/reference/rest/v1/documents/request#deletecontentrangerequest) where deleting content that contains a suggestion fails with this error:
```
googleapi: Error 500: Internal error encountered.
```
This affects all Hermes actions that rewrite the document header, including indexing updated documents, changing document state (e.g., requesting a review), and approving documents.

We have reported this to Google and they confirmed the bug, but there is no estimated time for a resolution. Unfortunately, while this bug exists, Hermes functionality to rewrite document headers fails if a user puts a suggestion in the header, and there is no way to automatically remediate this (working with suggestions is also unsupported by Google Docs' public API). Until this bug is fixed, we should lock documents in this state and provide some visual feedback for users for how to remediate, which is to remove all suggestions in the document header. Once all suggestions are removed from the document header, the Hermes indexer will automatically re-index the document and unlock it for editing.

PR #182, opened 2023-05-26
Lock documents in a bad state

jfreda:
Due to #181, we cannot refresh document headers when they contain a suggestion, so this PR locks documents in this state. The indexer will then always attempt to reindex locked documents on every run and unlock+reindex them if the suggestions in the header are removed. API calls that replace the document header have been updated to first check if the document is locked and will return a Locked HTTP status if this is the case.

For the frontend parts, this builds on @jeffdaley's work in #178 and updates the messaging a bit. Oh, and I couldn't take that `ReferenceError: error is not defined` message when requesting a review fails anymore, so I tracked that 🐛 down.

Bottom of sidebar when a document is locked:
image

### DIFF
--- a/internal/api/drafts.go
+++ b/internal/api/drafts.go
@@ -554,6 +554,26 @@ func DraftsDocumentHandler(
 			// Set custom editable fields.
 			docObj.SetCustomEditableFields()
 
+			// Get document from database.
+			doc := models.Document{
+				GoogleFileID: docId,
+			}
+			if err := doc.Get(db); err != nil {
+				l.Error("error getting document draft from database",
+					"error", err,
+					"path", r.URL.Path,
+					"method", r.Method,
+					"doc_id", docId,
+				)
+				http.Error(w, "Error requesting document draft",
+					http.StatusInternalServerError)
+				return
+			}
+
+			// Set locked value for response to value from the database (this value
+			// isn't stored in Algolia).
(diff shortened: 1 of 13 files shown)


jeffdaley (approved):

internal/api/drafts.go
```diff
@@ -695,6 +715,24 @@ func DraftsDocumentHandler(
...
+					"method", r.Method,
+					"path", r.URL.Path,
+					"doc_id", docId,
+				)
+				http.Error(w, "Error getting document status", http.StatusNotFound)
+				return
+			}
+			// Don't continue if document is locked.
+			if locked {
+				http.Error(w, "Document is locked", http.StatusLocked)
```
anubhavmishra, line 732: What happens in the frontend when we respond with this error? In some cases users may only notice that the document is locked when they change the state of the document. Do we have a good UX for them to identify what they should do to fix it? Can we point them to the sidebar where the locked message is shown?

cc: @jeffdaley

jfreda: Good questions! Currently this flash message will appear: 
image

It's not an ideal UX, but is better than the current experience of when a doc gets into this state 🙂  We should probably at least handle this error in a way where the sidebar immediately changes to "locked mode" (this currently requires a browser refresh). As we discussed offline, we can handle these in a follow-up PR.

anubhavmishra (approved):

PR #185, opened 2023-05-30
Add BadgeDropdownList component

jeffdaley:
Adds a generic BadgeDropdownList component. This will be the basis of the ProductSelect input in the document sidebar.

By default, uses the CheckableItem component:
CleanShot 2023-05-30 at 16 42 44@2x

But can yield any custom item:
CleanShot 2023-05-30 at 16 44 23@2x

### DIFF
--- a/web/app/components/inputs/badge-dropdown-list.hbs
+++ b/web/app/components/inputs/badge-dropdown-list.hbs
@@ -1,0 +1,51 @@
+{{!  @glint-nocheck - not typesafe yet }}
+<X::DropdownList
+  @items={{@items}}
+  @listIsOrdered={{@listIsOrdered}}
+  @onItemClick={{@onItemClick}}
+  @selected={{@selected}}
+  ...attributes
+>
+  <:anchor as |dd|>
+    <div class="relative w-full">
+      {{#if @isSaving}}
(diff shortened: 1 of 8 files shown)


web/app/components/inputs/badge-dropdown-list.hbs
```diff
@@ -0,0 +1,51 @@
...
+      {{/if}}
+      <dd.ToggleAction
+        data-test-badge-dropdown-trigger
+        class="relative {{if @isSaving 'opacity-50'}}"
+      >
+        <Hds::Badge
+          data-test-badge-dropdown-list-icon
+          data-test-icon={{if @selected (get-product-id @selected)}}
+          @text={{or @selected "--"}}
+          @icon={{if @selected (or (get-product-id @selected) "folder")}}
```
jfreda, line 27: If the intent is for this to be a generic component, should we make this less `product`-centric?

jfreda (approved):
One question, but not blocking 👍

web/app/components/inputs/badge-dropdown-list.hbs
jeffdaley: Good catch 🙏 Updated

PR #194, opened 2023-06-02
Fix post-auth redirect (Google auth, testing environment)

jeffdaley:
Fixes a bug where clicking the "Authenticate with Google" button would throw an assertion while in the local environment.

```
@tracked _info: AuthenticatedUser | null = null;

get info(): AuthenticatedUser {
  assert("Authenticated must exist", this._info);
  return this._info;
}
```

We were incorrectly checking for the getter rather than the tracked property.

### DIFF
--- a/web/app/services/session.ts
+++ b/web/app/services/session.ts
@@ -188,7 +188,7 @@ export default class SessionService extends EmberSimpleAuthSessionService {
   // Because we redirect as part of the authentication flow, the parameter storing the transition gets reset. Instead, we keep track of the redirectTarget in browser sessionStorage and override the handleAuthentication method as recommended by ember-simple-auth.
 
   handleAuthentication(routeAfterAuthentication: string) {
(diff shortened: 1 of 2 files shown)


web/app/services/session.ts
```diff
@@ -188,7 +188,7 @@ export default class SessionService extends EmberSimpleAuthSessionService {
   // Because we redirect as part of the authentication flow, the parameter storing the transition gets reset. Instead, we keep track of the redirectTarget in browser sessionStorage and override the handleAuthentication method as recommended by ember-simple-auth.
 
   handleAuthentication(routeAfterAuthentication: string) {
-    if (this.authenticatedUser.info) {
+    if (this.authenticatedUser._info) {
```
jeffdaley, line 191: This is actually the property that can be `null`. The non-underscored version is asserted true.

jfreda (approved):

PR #205, opened 2023-06-07
Add most type definitions for most HDS components

jeffdaley:
Adds mostly complete types for most of the HDS components we're using. For templates already type-checked by Glint, this removes many VSCode underlines and will allow make it much easier to remove `@glint-nocheck` declarations.

I generally skipped the component-like block yields for now, but hopefully we can make time to add them soon.

### DIFF
--- a/web/app/components/modals/doc-created.hbs
+++ b/web/app/components/modals/doc-created.hbs
@@ -1,4 +1,3 @@
-{{! @glint-nocheck: not typesafe yet }}
 {{#if this.modalIsShown}}
   <Hds::Modal data-test-doc-created-modal as |M|>
     <M.Header>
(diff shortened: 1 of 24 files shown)


web/app/components/modals/doc-created.hbs
```diff
@@ -17,7 +16,7 @@
         <Hds::Link::Inline
           @icon="external-link"
           @iconPosition="trailing"
-          @hrefIsExternal={{true}}
+          @isHrefExternal={{true}}
```
jeffdaley, line 19: Glint doing its job. (I'm sure I'll find more mistakes as we remove `nocheck`s.

jfreda (approved):

PR #206, opened 2023-06-08
Type Sidebar

jeffdaley:
[Finally] adds types to the Sidebar component.

### DIFF
--- a/web/app/components/document/sidebar.hbs
+++ b/web/app/components/document/sidebar.hbs
@@ -313,64 +313,8 @@
       </div>
     </div>
 
-    {{!-- <div class="flex flex-col items-start space-y-2 px-3">
-      <small class="hds-typography-body-100 hds-foreground-faint">Tags</small>
-      {{#if this.isOwner}}
-        <EditableField
-          @value={{this.tags}}
-          @onChange={{perform this.save "tags"}}
-          @loading={{this.save.isRunning}}
-        >
-          <:default>
-            {{#if this.tags.length}}
-              <ol>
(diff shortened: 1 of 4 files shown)


web/app/components/document/sidebar.hbs
```diff
@@ -313,64 +313,8 @@
...
-                {{tag}}
-              </li>
-            {{/each}}
-          </ol>
-        {{else}}
-          <em>No tags</em>
-        {{/if}}
-      {{/if}}
-    </div> --}}
-
```
jeffdaley, line 368: It's time. (Nothing here we can't easily replicate later if we want to implement tags.)

web/app/types/document.d.ts
```diff
@@ -12,8 +14,16 @@ export interface HermesDocument {
   docNumber: string;
   docType: string;
   title: string;
+  locked?: boolean;
+  owners?: string[];
+  appCreated?: boolean;
+  contributors?: HermesUser[];
+  approvers?: HermesUser[];
+  changesRequestedBy?: unknown[]; // TODO: Define this
```
jfreda, line 22: This should also be `HermesUser[]`

jfreda (approved):
🙌

web/app/types/document.d.ts
jeffdaley: Actually I'm inferring that it's a `string[]` like `approvedBy`:

```
  get hasRequestedChanges() {
    return this.args.document.changesRequestedBy?.includes(
      this.args.profile.email
    );
  }
```

PR #207, opened 2023-06-08
Type Doc components

jeffdaley:
Adds types to the Doc components.


web/app/components/doc/inline.hbs
```diff
@@ -18,8 +19,10 @@
 
     <Person @ignoreUnknown={{true}} @imgURL={{@avatar}} @email={{@owner}} />
 
-    {{#if (and @isResult @snippet)}}
```
jeffdaley, line 21: Need to investigate/report this. Glint thought `@snippet` was undefined despite it passing the `and` check.

web/app/components/doc/row.ts
```diff
@@ -0,0 +1,75 @@
...
+  Args: {
+    productArea: string;
+  };
+}
+
+export default class DocRowComponent extends Component<DocRowComponentSignature> {
+  get productAreaName() {
+    let productAreaName;
+
+    // TODO: Convert this to a switch statement
```
jfreda, line 13: I thought we had a helper for this logic somewhere 🤔

jfreda (approved):

web/app/components/doc/row.ts
jeffdaley: Ah right, yeah we have one for the icon.

This one just needs to be cleaned up (and, separately, turned into its own helper).

PR #210, opened 2023-06-09
Fix a couple ProductAreaDropdown bugs

jeffdaley:
- Adds validation to the form when you select a product (i.e., the "Create Doc on Google" enables on ProductAreaSelect)
- Saves the `productAbbreviation` when creating a new doc (by way of new `attributes` support in `X::DropdownList`
- Improves the revert-on-error logic

### DIFF
--- a/web/app/components/document/sidebar.ts
+++ b/web/app/components/document/sidebar.ts
@@ -236,29 +236,31 @@ export default class DocumentSidebarComponent extends Component<DocumentSidebarC
     });
   }
 
-  updateProduct = restartableTask(async (product: string) => {
-    this.product = product;
-    await this.save.perform("product", this.product);
-  });
+  updateProduct = restartableTask(
+    async (product: string, productAbbreviation: string) => {
+      this.product = product;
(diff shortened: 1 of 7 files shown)


web/app/components/document/sidebar.ts
```diff
@@ -236,29 +236,32 @@ export default class DocumentSidebarComponent extends Component<DocumentSidebarC
...
+      try {
         await this.patchDocument.perform({
-          [field]: cleanString(serializedValue),
+          [field]: serializedValue,
         });
       } catch (err) {
         // revert field value on failure
-        targetField = oldVal;
+        (this as any)[field] = val;
+        console.error(err);
```
jfreda, line 264: Did you mean to leave this in or was it just for debugging locally?

jfreda (approved):

web/app/components/document/sidebar.ts
jeffdaley: Intentional, but it makes more sense as a flashMessage. PR'd here: https://github.com/hashicorp-forge/hermes/pull/211

PR #212, opened 2023-06-13
Lock documents on other types of suggestions in the header

jfreda:
Builds on #182 to lock documents when there are other types of suggestions in the header (which causes the same issue: #181).

### DIFF
--- a/internal/indexer/refresh_headers.go
+++ b/internal/indexer/refresh_headers.go
@@ -89,7 +89,11 @@ func refreshDocumentHeaders(
 		}
 		lockedDocIDs = append(lockedDocIDs, d.GoogleFileID)
 	}
-	log.Info(fmt.Sprintf("locked document IDs: %v", lockedDocIDs))
+	if ft == draftsFolderType {
+		log.Info(fmt.Sprintf("locked draft document IDs: %v", lockedDocIDs))
+	} else {
+		log.Info(fmt.Sprintf("locked document IDs: %v", lockedDocIDs))
+	}
 
 	// Return if there are no updated documents.
 	if len(docs) == 0 {
--- a/pkg/hashicorpdocs/locked.go
+++ b/pkg/hashicorpdocs/locked.go
@@ -104,15 +104,37 @@ func containsSuggestionInHeader(doc *docs.Document) bool {
 
 	// Navigate through all table contents to look for suggestions.
 	for _, row := range t.TableRows {
+		if len(row.SuggestedDeletionIds) > 0 ||
+			len(row.SuggestedInsertionIds) > 0 ||
+			len(row.SuggestedTableRowStyleChanges) > 0 {
+			// We found a suggestion.
+			return true
+		}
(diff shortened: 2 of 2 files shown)


pkg/hashicorpdocs/locked.go
```diff
@@ -104,15 +104,37 @@ func containsSuggestionInHeader(doc *docs.Document) bool {
...
 	// Navigate through all table contents to look for suggestions.
 	for _, row := range t.TableRows {
+		if len(row.SuggestedDeletionIds) > 0 ||
+			len(row.SuggestedInsertionIds) > 0 ||
+			len(row.SuggestedTableRowStyleChanges) > 0 {
+			// We found a suggestion.
+			return true
+		}
 		for _, cell := range row.TableCells {
+			if len(cell.SuggestedDeletionIds) > 0 ||
```
anubhavmishra, line 114: Could we add a few comments about what is being checked here and what level are we in wrt to the header table?

pkg/hashicorpdocs/locked.go
```diff
@@ -104,15 +104,37 @@ func containsSuggestionInHeader(doc *docs.Document) bool {
...
+				len(cell.SuggestedInsertionIds) > 0 ||
+				len(cell.SuggestedTableCellStyleChanges) > 0 {
+				return true
+			}
 			for _, content := range cell.Content {
 				if para := content.Paragraph; para != nil {
+					if len(para.SuggestedBulletChanges) > 0 ||
```
anubhavmishra, line 121: Could we add a few comments about what is being checked here and what level are we in wrt to the header table?

pkg/hashicorpdocs/locked.go
```diff
@@ -104,15 +104,37 @@ func containsSuggestionInHeader(doc *docs.Document) bool {
...
+						len(para.SuggestedParagraphStyleChanges) > 0 ||
+						len(para.SuggestedPositionedObjectIds) > 0 {
+						return true
+					}
 					for _, elem := range para.Elements {
+						if auto := elem.AutoText; auto != nil {
```
anubhavmishra, line 127: Could we add a few comments about what is being checked here and what level are we in wrt to the header table?

internal/indexer/refresh_headers.go
```diff
@@ -89,7 +89,11 @@ func refreshDocumentHeaders(
 		}
 		lockedDocIDs = append(lockedDocIDs, d.GoogleFileID)
 	}
-	log.Info(fmt.Sprintf("locked document IDs: %v", lockedDocIDs))
+	if ft == draftsFolderType {
+		log.Info(fmt.Sprintf("locked draft document IDs: %v", lockedDocIDs))
+	} else {
+		log.Info(fmt.Sprintf("locked document IDs: %v", lockedDocIDs))
+	}
```
anubhavmishra, line 96: 🎉

anubhavmishra (approved):

PR #219, opened 2023-06-15
Related resources

jeffdaley:
Adds a `RelatedResources` component, allowing editors to attach first- and third-party links to their document.


jfreda (commented):
Still reviewing, just a couple questions so far...

web/app/components/document/sidebar/related-resources.ts
```diff
@@ -0,0 +1,459 @@
...
+              "Content-Type": "application/json",
+            },
+          }
+        );
+      } catch (e: unknown) {
+        this.relatedLinks = cachedLinks;
+        this.relatedDocuments = cachedDocuments;
+
+        this.flashMessages.add({
+          title: "Unable to save",
```
jfreda, line 444: Nit: might be somewhat helpful for the user to differentiate between another save (doc?) action.
```suggestion
          title: "Unable to save related resources",
```

web/app/components/document/sidebar/related-resources/add.ts
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * The action to enable keyboard navigation.
+   * Called when the search input receives focus.
+   */
+  @action protected enableKeyboardNav() {
+    this.keyboardNavIsEnabled = true;
+  }
+
+  @action protected onItemClick(_item: any, attrs: any) {
```
jfreda, line 159: No biggie, but just noticed that this was the only action without a comment, in case it was an oversight.

web/app/helpers/get-link-icon.ts
```diff
@@ -0,0 +1,54 @@
...
+  Args: {
+    Positional: [string | undefined];
+  };
+  Return: string;
+}
+
+const getLinkIconHelper = helper<GetLinkIconSignature>(([url]) => {
+  if (url) {
+    const urlParts = url.split("/");
+    let domain = urlParts[2];
```
jfreda, line 14: Should we have more URL validation at this level? What kind of UX is it if we fail the `assert` below?

web/app/components/document/sidebar/related-resources/add.ts
```diff
@@ -0,0 +1,292 @@
...
+  @action protected onExternalLinkTitleInput(e: Event) {
+    const input = e.target as HTMLInputElement;
+    this.externalLinkTitle = input.value;
+  }
+
+  /**
+   * The action to check for duplicate ExternalResources.
+   * Used to dictate whether a warning message is displayed.
+   */
+  @action private checkForDuplicate(url: string) {
```
jfreda, line 184: We're not actually enforcing this in the backend, but this is a good idea.

jfreda (approved):
Consider the resources related (just some non-blocking questions/nits).

web/app/components/document/sidebar/related-resources.ts
```diff
@@ -0,0 +1,459 @@
...
+          filters: filterString,
+          attributesToRetrieve: [
+            "title",
+            "product",
+            "docNumber",
+            "docType",
+            "status",
+            "owners",
+          ],
+          // TODO: Confirm this with a fresh index
```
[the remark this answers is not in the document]
jeffdaley: I think we can remove it but let's make sure to test this on staging.

I have Algolia index problems that prevent me from trying this locally.

web/app/components/document/sidebar/related-resources.ts
jeffdaley: I kept this generic so it could also apply to non "Related resources"-labeled fields, e.g., "RFC." I've changed it to "Unable to save resource" which might split the difference? What do you think

PR #231, opened 2023-06-22
Rename some Dropdown-related properties

jeffdaley:
Changes the name of some `selected` properties to `isSelected` to differentiate them from other `selected` properties. 

In the context of DropdownLists, we use `selected` in two ways: 

1. "Among a list, which item(s) [are] selected?"
2. "Is this item selected?"

It's hard to remember which one is which, so I'm renaming the latter cases.

Here's a quick visualization of the changes:
CleanShot 2023-06-22 at 10 01 34@2x


jfreda (approved):
Makes sense!

PR #232, opened 2023-06-23
Add EmberAnimatedTools

jeffdaley:


web/config/environment.js
```diff
@@ -66,7 +66,9 @@ module.exports = function (environment) {
       },
     },
 
-    featureFlags: {},
+    featureFlags: {
+      showEmberAnimatedTools: getEnv("SHOW_EMBER_ANIMATED_TOOLS", false),
+    },
```
jfreda, line 71: I think this probably fits more as its own config setting instead of a feature flag.

```suggestion
    showEmberAnimatedTools: getEnv("SHOW_EMBER_ANIMATED_TOOLS", false),

    featureFlags: {},
```

web/app/controllers/application.ts
```diff
@@ -0,0 +1,16 @@
...
+import { inject as service } from "@ember/service";
+import ConfigService from "hermes/services/config";
+import config from "hermes/config/environment";
+
+export default class ApplicationController extends Controller {
+  @service("config") declare configSvc: ConfigService;
+
+  protected get animatedToolsAreShown() {
+    if (config.environment === "development") {
+      return this.configSvc.config.feature_flags["showEmberAnimatedTools"];
```
jfreda, line 11: I _think_ this should work along with the other suggestion I made below...
```suggestion
import config from "hermes/config/environment";

export default class ApplicationController extends Controller {
  protected get animatedToolsAreShown() {
    if (config.environment === "development") {
      return config.showEmberAnimatedTools;
```

jfreda (approved):

PR #237, opened 2023-06-26
Add new configurations to FloatingUI component

jeffdaley:
- Adds a `null` placement option to ignore dynamic positioning.
- Adds a `disableClose` argument

These are admittedly roundabout ways of extending the DropdownList components to work in non-popover contexts. I intend to improve this in future iterations, after the dust of #219 has settled.

### DIFF
--- a/web/app/components/floating-u-i/content.ts
+++ b/web/app/components/floating-u-i/content.ts
@@ -18,6 +18,7 @@ interface FloatingUIContentSignature {
   Args: {
     anchor: HTMLElement;
     id: string;
+    // TODO: Move "none" logic to a parent component.
     placement?: Placement | "none";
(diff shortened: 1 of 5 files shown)


web/app/components/floating-u-i/content.ts
```diff
@@ -18,7 +18,8 @@ interface FloatingUIContentSignature {
   Args: {
     anchor: HTMLElement;
     id: string;
-    placement?: Placement;
+    // TODO: Move "none" logic to a parent component.
+    placement?: Placement | "none";
```
jfreda, line 22: Curious why we use a "none" string here instead of `null`?

jfreda (approved):
float on

PR #240, opened 2023-06-27
EditableField types and tests

jeffdaley:
This is the broader result of the #243 bug: more precisely typed and tested CustomEditableField components.


web/app/components/document/sidebar.hbs
```diff
@@ -288,7 +288,6 @@
                 @field={{field}}
                 @attributes={{attributes}}
                 @onChange={{perform this.save field}}
-                @updateFieldValue={{this.updateCustomFieldValue}}
```
jeffdaley, line 291: This wasn't being used (anymore?) as far as I could tell.

web/app/components/editable-field.ts
```diff
@@ -1,64 +1,92 @@
...
 import { action } from "@ember/object";
 import { scheduleOnce } from "@ember/runloop";
+import { assert } from "@ember/debug";
 
 export const FOCUSABLE =
   'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
 
-export default class EditableField extends Component {
-  @tracked editing = false;
-  @tracked element = null;
```
jeffdaley, line 13: `element` is a protected propertyName according to Glint/TS so I changed it to `el`

jfreda (approved):

PR #273, opened 2023-07-26
(WIP) Add support screen

jeffdaley:
Adds a support screen that's accessible from the user menu. It reads JSON/Markdown from [somewhere] and formats it in the template.


jeffdaley:
Closing in favor of #275

PR #280, opened 2023-08-08
Delete hermes-urls.d.ts

jeffdaley:
In an e

### DIFF
--- a/web/app/utils/hermes-urls.d.ts
+++ b/web/app/utils/hermes-urls.d.ts
@@ -1,5 +1,0 @@
-declare module "hermes/utils/hermes-urls" {
-  export const HERMES_GITHUB_REPO_URL: string;
-  export const TEST_SUPPORT_URL: string;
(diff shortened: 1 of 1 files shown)


web/app/utils/hermes-urls.d.ts
```diff
@@ -1,5 +0,0 @@
-declare module "hermes/utils/hermes-urls" {
-  export const HERMES_GITHUB_REPO_URL: string;
-  export const TEST_SUPPORT_URL: string;
-  export const TEST_SHORT_LINK_BASE_URL: string;
-}
```
jeffdaley, line 5: I shouldn't have been trying to export `const`s from a `d.ts` in the first place.

jeffdaley: I made this to appease Mirage but it turns out not to need it.

jfreda (approved):

PR #294, opened 2023-08-15
Add `title` attribute to RelatedResources

jeffdaley:
Adds `title` attribute to Related Resources so they can be hovered over for the full title. 

+ Prettier className reorganization


web/app/components/document/sidebar/related-resources/list-item/resource.hbs
```diff
@@ -16,12 +16,13 @@
 <div class="w-full overflow-hidden">
   <TruncatedText
     @tagName="h4"
-    class="text-body-200 text-color-foreground-strong font-medium"
+    class="text-body-200 font-medium text-color-foreground-strong"
+    title={{this.title}}
```
jeffdaley, line 20: This is the non-Prettier change

anubhavmishra (approved):

PR #297, opened 2023-08-17
Add version and revision in config response

anubhavmishra:
This PR adds `version` and `short_revision` in the `/api/v1/web/config` API response.

Note: This change won't affect the `hermes version` command output. It will continue showing the version number of the application.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -1,3 +1,31 @@
 package version
 
+import (
+	"runtime/debug"
+)
+
 const Version = "0.2.0"
+
+// GetVersion returns
+// the version number
+func GetVersion() string {
+	return Version
+}
+
+// GetShortRevision returns a short
+// commit or checkout revision
+// from build information
+func GetShortRevision() string {
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" {
+				var shortRevision string
+				if len(setting.Value) > 7 {
+					shortRevision = setting.Value[:7]
+				}
+				return shortRevision
+			}
+		}
+	}
+	return ""
+}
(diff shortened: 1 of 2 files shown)


internal/version/version.go
```diff
@@ -1,3 +1,31 @@
...
+// the version number
+func GetVersion() string {
+	return Version
+}
+
+// GetShortRevision returns a short
+// commit or checkout revision
+// from build information
+func GetShortRevision() string {
+	if info, ok := debug.ReadBuildInfo(); ok {
```
anubhavmishra, line 19: Go `1.18` allow for version control information. https://pkg.go.dev/runtime/debug#ReadBuildInfo

jeffdaley:
LGTM! Is revision the same as SHA? If so, my preference would be to call it a SHA, since that's what we've been referring to it internally as. But I'm also happy to learn a new term.

Aside: TIL 7-char SHAs redirect to the full URL.

jeffdaley (approved):

anubhavmishra:
@jeffdaley Yea, I thought about this and settled on revision. Wanted to follow what Golang suggests `vcs.revision` given VCS could be any version control system.

internal/version/version.go
```diff
@@ -1,3 +1,31 @@
...
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" {
+				var shortRevision string
+				if len(setting.Value) > 7 {
+					shortRevision = setting.Value[:7]
```
jfreda, line 24: It looks like this will return an empty string if the value length is >= 7. Do we want to do something like this to be safer and always return the value?

```suggestion
				shortRevision := setting.Value
				if len(shortRevision) > 7 {
					shortRevision = setting.Value[:7]
```

anubhavmishra: Good catch!

jfreda (approved):

PR #299, opened 2023-08-18
Add support for single-file components

jeffdaley:
Adds support for template imports, i.e., single-file components, and converts `action.hbs/ts` to `action.gts`.

Single-file components are are expected to be the recommended authoring format for Ember 5x; might as well get good seats on that train (source: [Ember Blog](https://blog.emberjs.com/ember-5-0-released/)).

Guided by [▶ EmberConf 2023 - Getting Started with 'template' Tag Components by Ignace Maes](https://www.youtube.com/watch?v=kjMo0MaJF1Q) which also does a good job showing the benefits of this new format, such as locally defined components (shown below). It should boost DX, reduce file count, and make code review easier.

```
const MyListItem =
  <template>
    <div>{{yield}}</div>
  </template>;

const MyList =
  <template>
    <div>List of things</div>
      {{#each @items as |item|}}
        <MyListItem>
          {{item.number}}-{{item.value}}
        </MyListItem>
      {{/each}}
  </template>;

export default MyList;
```


web/app/components/action.gts
```diff
@@ -7,7 +7,13 @@ interface ActionComponentSignature {
...
 }
 
-export default class ActionComponent extends Component<ActionComponentSignature> {}
+export default class ActionComponent extends Component<ActionComponentSignature> {
+  <template>
+    <button type="button" class="action" ...attributes>
+      {{yield}}
+    </button>
+  </template>
+}
```
jeffdaley, line 16: The short-term downside of `.gts` is that GitHub doesn't have syntax highlighting yet.

jfreda: Looks like we should be able to configure this using an [override](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md) in a `.gitattributes` file 🙌

jfreda (approved):
Cool, let's give it a try! 🚀

PR #303, opened 2023-08-22
ExternalResource: New icon and URL

jeffdaley:
- Changes the default external-link icon to `globe` (the `file` icon had an "image didn't load" feel to it whereas the globe feels more deliberate).
- Adds a short URL to external resources to make the link target clearer.

Shown here with a related doc (unchanged) for reference:
CleanShot 2023-08-22 at 14 50 11@2x

### DIFF
--- a/web/app/styles/components/document/related-resources.scss
+++ b/web/app/styles/components/document/related-resources.scss
@@ -1,8 +1,8 @@
 .related-resources-list {
-  @apply w-full mt-0.5;
+  @apply mt-0.5 w-full;
 
   .related-resource-inner {
-    @apply w-full relative;
+    @apply relative w-full;
 
     &:hover,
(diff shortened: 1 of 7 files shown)


web/app/styles/components/document/related-resources.scss
```diff
@@ -89,30 +89,24 @@
...
 
   .hermes-document + .external-resource {
     @apply mt-1.5;
   }
-
-  .external-resource {
-    .related-resource-link {
-      @apply items-center;
-    }
-  }
```
jeffdaley, line 111: This is the actual change. The rest is Prettier (apologies).

jfreda: 💅

jfreda (approved):

PR #304, opened 2023-08-23
Fix recently viewed docs disappearing bug

jfreda:
Recently viewed docs can disappear for a user after updating email notifications 😬 This PR fixes the bug.

### DIFF
--- a/internal/api/me_subscriptions.go
+++ b/internal/api/me_subscriptions.go
@@ -93,17 +93,28 @@ func MeSubscriptionsHandler(
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


internal/api/me_subscriptions.go
```diff
@@ -93,17 +93,28 @@ func MeSubscriptionsHandler(
 				return
 			}
 
+			// Find or create user.
+			u := models.User{
```
anubhavmishra, line 97: 💯

anubhavmishra (approved):

PR #306, opened 2023-08-24
Change `/all` to `/documents`

jeffdaley:
Moves the "All Docs" route from `/all` to `/documents` to give us a better framework to add new models, e.g., projects, products, teams, that can be browsed. References to the `/all` route are now redirected to `/documents`.


web/app/controllers/authenticated/all.ts
```diff
@@ -1,8 +1,4 @@
 import Controller from "@ember/controller";
```
jfreda, line 1: Do we need this file anymore?

jfreda (approved):
Do you think we should add an acceptance test for the redirect? (not sure offhand if this is possible, but if so, it might be able to catch regressions)

web/app/controllers/authenticated/all.ts
jeffdaley: It's needed to pass the queryParams to the `/documents` route on redirect. I've added a note to the controller.

jeffdaley:
> Do you think we should add an acceptance test for the redirect?

Thanks for catching this. I wrote one in my spike branch but forgot to bring it over. ✅

PR #313, opened 2023-08-28
Replace RelatedResource FlightIcons with Favicons

jeffdaley:
Replaces ExternalResource FlightIcons with Google-fetched favicons. Falls back to a generic icon on error.


web/app/components/favicon.hbs
```diff
@@ -0,0 +1,6 @@
+<img
+  data-test-favicon
+  src="https://www.google.com/s2/favicons?domain={{@url}}&sz=32"
```
jfreda, line 3: I feel a bit uneasy about only using an unofficial third-party API for this, given this could seemingly just disappear at any point and then break all icons. What do you think about only making this a fallback if we don't have a flight icon for the URL?

web/app/components/document/sidebar/related-resources/list-item/resource.hbs
```diff
@@ -6,10 +6,7 @@
       <img src="/images/document.png" class="h-auto w-[18px]" />
     </div>
   {{else}}
-    <FlightIcon
-      @name={{get-link-icon this.url}}
-      class={{if (eq (get-link-icon this.url) "globe") "opacity-40"}}
-    />
+    <Favicon @url={{this.url}} />
```
jfreda, line 9: What's the UX if the URL is malformed/invalid?

jeffdaley: Google responds with a default image (globe icon, 16x16)

web/app/components/favicon.hbs
jeffdaley: Yeah I share your concern so some extent. I'll shift this PR back to draft while I bulletproof it.

web/app/components/favicon.hbs
```diff
@@ -0,0 +1,17 @@
+{{#if this.fallbackIconIsShown}}
+  <FlightIcon
+    ...attributes
+    data-test-fallback-favicon
+    @name="globe"
```
jfreda, line 5: Should we keep the logic in `get-link-icon.ts` (that's currently deleted in this PR) to get a more specific icon?

jeffdaley: It seemed more confusing than helpful in the fallback scenario, where the job is to communicate "something failed to load." If we're starting with three favicon'd links, e.g.,

🟧 Asana discussion
🌈 Figma prototype
🟦 Algolia docs

and then one day it looks like

🌐 Asana discussion
🌈 Figma prototype
🌐 Algolia docs

jfreda: Got it, makes sense 👍

jfreda (approved):

PR #316, opened 2023-08-31
Add ProductArea service

jeffdaley:
Adds a ProductArea service to fetch and store product areas.

Currently these functions live in the `ProductSelect` component, so there's no caching, and state is reset on every render. This means we're fetching and loading more than we should. With a service, we can fetch the list once and it'll persist the session.

I'll soon use this service to add `productArea` to the `/new` `model`, which should improve that route's UX.

**Observe the problem on `main`**
- Create a RFC and watch for the ProductArea dropdown's loading state. (You can refresh if you miss it.)
- Click the back button
- Click the RFC tile again
- Notice the ProductArea dropdown's unnecessary loading state

### DIFF
--- a/web/app/components/document/sidebar/related-resources.hbs
+++ b/web/app/components/document/sidebar/related-resources.hbs
@@ -13,7 +13,7 @@
   </div>
 {{else if this.loadingHasFailed}}
   <div class="h-16">
-    <div class="related-resources-failed-to-load">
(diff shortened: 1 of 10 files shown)


jfreda (approved):
Nice - we discussed offline about perhaps exploring using Ember Data instead of a service per endpoint like this, but this seems like a good improvement to the current implementation 👍

PR #317, opened 2023-08-31
Adds `matchAnchorWidth` to FloatingUI components

jeffdaley:
Adds `matchAnchorWidth` support to FloatingUI components to set popover widths relative to their anchors. The basic invocation is `@matchAnchorWidth={{true}}` but you can also pass in a `{{hash}}` with `enabled` and `additionalWidth` properties for more control. (See the Search component for an example.)


jfreda (approved):
Nice

PR #318, opened 2023-09-01
Improve RecentlyViewedDocs/Dashboard performance

jeffdaley:
Improves performance of the RecentlyViewedDocs service and Dashboard load times.

To test:
1. Load the Hermes Dashboard
2. Click to another route
3. Return to the Dashboard
4. Compare performance vs. main

### DIFF
--- a/web/app/routes/authenticated/dashboard.ts
+++ b/web/app/routes/authenticated/dashboard.ts
@@ -61,13 +61,13 @@ export default class DashboardRoute extends Route {
      *
      * If a user leaves and returns to the dashboard, we call `fetchAll`
      * but don't await it. This speeds up the transition and is especially
-     * efficient when the RecentDocs list hasn't changed since last load.
+     * efficient when the RecentDocs list hasn't changed since its last load.
      *
      * It's possible for the user to see the RecentDocs list update
      * in real time (by visiting a document and very quickly clicking back),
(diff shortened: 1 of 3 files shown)


web/app/routes/authenticated/dashboard.ts
```diff
@@ -54,13 +54,26 @@ export default class DashboardRoute extends Route {
...
+     */
+    if (this.recentDocs.all) {
+      void this.recentDocs.fetchAll.perform();
+    } else {
       await this.recentDocs.fetchAll.perform();
-    } catch {
-      /**
-       * This tells our template to show the error state.
-       */
-      this.recentDocs.all = null;
```
jfreda, line 63: Do we still want this in case of an error when fetching recent docs?

jeffdaley: This is now in the service: https://github.com/hashicorp-forge/hermes/pull/318/files#diff-e4377cd5d1c050300d977e533d7234362dbf38040020709bc2188ca12951bf3bR102

And should still trigger the template error: https://github.com/hashicorp-forge/hermes/blob/main/web/app/templates/authenticated/dashboard.hbs#L49

jfreda: Ah great 👍

jfreda (approved):
Good catch!

PR #321, opened 2023-09-05
Update HDS/Flight Icon packages

jeffdaley:
Updates the HDS packages.

On the `design-system-components` side, we gain a fix to the modal's `isDismissedDisabled` argument: Users will no longer be able to dismiss modals while the primary task is running (https://github.com/hashicorp/design-system/pull/1456). In FlightIcon world, we now have the Jira logo.

I did a pretty thorough clickthrough and caught only 1 visual diff. It looks like there's a collision between HDS's and our "tooltip.scss" files, so I renamed ours to `x-tooltip.scss`.


jfreda (approved):
📈

PR #323, opened 2023-09-06
Replace `modifiedAgo` property with template helper

jeffdaley:
Replaces the computed `modifiedAgo` property from the Document model with template-calculated equivalents.

For now, this feels more ergonomic and Ember-like (until we use Ember Data).


jfreda (approved):
Nice

PR #327, opened 2023-09-07
Related Resources refactor and External Link `scope`

jeffdaley:
Refactors and expands the RelatedResources component so it can be used outside of the DocumentSidebar context/format.

- Moves potentially shared types and functions (most of what happens in-modal) from `/components/documents/sidebar` → `/components`
- Adds an external-link `scope` to skip remove the "search Algolia" functionality. (There's also a document scope that replaces `allowAddingExternalLinks=false`), 
- Renames some external-link-related files and elements differentiate between `scoped` and `fallback` cases


web/app/components/document/sidebar.hbs
```diff
@@ -299,7 +299,6 @@
           @documentIsDraft={{this.isDraft}}
           @productArea={{@document.product}}
           @objectID={{@document.objectID}}
-          @allowAddingExternalLinks={{true}}
```
jeffdaley, line 302: This is now the default setting.

web/app/components/related-resources/add-or-edit-external-resource-modal.ts
```diff
@@ -89,44 +100,46 @@ export default class DocumentSidebarRelatedResourcesListItemEditComponent extend
...
 
     if (this.urlIsValid) {
-      this.args.onSave(newResource);
+      const sortOrder = this.args.resource ? this.args.resource.sortOrder : 1;
+
+      this.args.onSave({
+        sortOrder,
+        name: this.title,
+        url: this.url,
+      });
```
jeffdaley, line 136: Note to self: Investigate as potential cause of EmberAnimated bug

web/app/components/document/sidebar/related-resources/list.ts
```diff
@@ -71,6 +71,17 @@ export default class DocumentSidebarRelatedResourcesListComponent extends Compon
...
+      /**
+       * For unknown reasons, Ember Animated sees an updated sprite
+       * as both inserted and removed, causing unwanted animations.
+       * This check prevents that from happening.
+       *
+       * FIXME: Investigate and fix this.
+       */
+      return;
+    }
+
```
jeffdaley, line 84: A rather disappointing side effect of the refactor: Recently edited sprites are no longer interpreted by Ember Animated as `kept` but rather `removed` _and_ `inserted`.

I have some theories that I'll investigate soon.

web/app/components/document/sidebar/related-resources/list.hbs
```diff
@@ -6,6 +6,8 @@
       </div>
     {{else}}
       <ul class="related-resources-list" ...attributes>
+        {{! This is being looped through twice for some reason... }}
+        {{! On immediate save, the order is wrong - maybe an array push thing? }}
```
jfreda, line 10: Should we follow up on this (/add a TODO)?

jfreda (approved):

web/app/components/document/sidebar/related-resources/list.hbs
jeffdaley: This was actually resolved and I forgot to remove the comment. Thanks for catching.

PR #329, opened 2023-09-08
Remove unused Avatar styles

jeffdaley:
Removes unused border properties to make Avatars look crisper.


web/app/components/person/avatar.hbs
```diff
@@ -1,14 +1,14 @@
 <div
-  class="overflow-hidden rounded-full flex justify-center items-center bg-color-palette-alpha-300 bg-color-palette-neutral-200 border border-transparent
```
jeffdaley, line 2: The actual, non-Prettier change is removing `border border-transparent`

PR #333, opened 2023-09-08
Autofocus PeopleSelects in the sidebar

jeffdaley:
Fixes a bug where clicking an EditableField+PeopleSelect wouldn't focus it as expected.


web/app/components/inputs/people-select.ts
```diff
@@ -17,6 +17,8 @@ interface InputsPeopleSelectComponentSignature {
   Args: {
     selected: HermesUser[];
     onChange: (people: HermesUser[]) => void;
+    renderInPlace?: boolean;
+    disabled?: boolean;
```
jeffdaley, line 21: Happened to come across this missing types. SORRY

PR #335, opened 2023-09-11
Improve document route performance

jeffdaley:
Improves performance of the `/document` route by moving non-essential tasks out of the `model`.

One-way tasks, i.e., those that don't affect local properties, are now `void`-called in the `afterModel` hook. User avatars are loaded by a new `<HermesUsers>` component that returns avatar `imgURL`s from an array of emails (including Stakeholders and any other `PEOPLE`-type CustomEditableField). `docType` continues to be a model attribute, but we no longer await it. Instead, we provide the promise to the component to be awaited contextually.

Plus:
- Tweaks the EditableField API to differentiate between `isLoading` and `isSaving` states.
- Improves CSS to minimize layout shift when loading


jfreda (approved):
📈

PR #346, opened 2023-09-25
Update notification width and `out` animation

jeffdaley:
Fixes a couple FlashNotification bugs:

- Notifications are now fixed width rather than `fit-content` avoiding a case where stacked notifications would be different -widths:
CleanShot 2023-09-25 at 09 40 15@2x

- The `notificationOut` animation now works as expected. Currently you have to mouseover the div to trigger it.

### DIFF
--- a/web/app/styles/components/notification.scss
+++ b/web/app/styles/components/notification.scss
@@ -13,12 +13,29 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


jfreda (approved):
Good catch 👍

PR #349, opened 2023-09-26
Add `secondaryFilterAttribute` to DropdownList

jeffdaley:
Adds a `secondaryFilterAttribute` argument to the DropdownList so items can be filtered by additional information. For example, in the ProductSelect, we now match abbreviation in addition to productAreaName:  

CleanShot 2023-09-26 at 16 20 42@2x

*Avatar designs placeholder


jfreda (approved):
Nice improvement

PR #366, opened 2023-10-11
Update "Choose a template" screen

jeffdaley:
- Updates the "Choose a template" screen to be prettier, tidier and more space efficient.
- Collects `moreInfoLinks`, if they exist, in a section below the templates.
- Adds DocType icons and a hook for user-customizable icons
- Adds Mirage factory for document types and cleans up the default response
- Adds tests
- Includes commented-out styles for the projects feature


web/app/helpers/get-doctype-icon.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+  Return: string;
+}
+
+/**
+ * Returns the FlightIcon name for a given document type.
+ * This is a temporary helper until "flight_icon" is added
+ * to the HermesDocumentType model.
+ */
+const getDoctypeIconHelper = helper<GetDoctypeIconSignature>(
```
jfreda, line 15: We discussed this offline, but we'll move this to the document types API soon.

jfreda (approved):

PR #367, opened 2023-10-12
Simplify application templating

jeffdaley:
Moves shared templating to `authenticated.hbs` and moves the `Toolbar` out of the `Header`. Basically reduces the number of times we're writing `<div class="x-container">` and `<Header />`.


web/app/components/footer.hbs
```diff
@@ -1,38 +1,34 @@
-{{#unless (eq this.currentRouteName "authenticated.document")}}
```
jeffdaley, line 1: This check is now in `authenticated.ts`

jfreda (approved):

PR #370, opened 2023-10-18
RelatedResources: Fix `documentType` bug

jeffdaley:
Fixes a bug preventing a `RelatedHermesDocument`'s `documentType` from saving, and similarly fixes the Mirage type that was incorrectly giving us a passing test.

### DIFF
--- a/web/app/components/related-resources/add.ts
+++ b/web/app/components/related-resources/add.ts
@@ -331,9 +331,6 @@ export default class RelatedResourcesAddComponent extends Component<RelatedResou
       title: attrs.title,
       documentType: attrs.docType,
       documentNumber: attrs.docNumber,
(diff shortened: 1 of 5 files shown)


web/app/components/related-resources/add.ts
```diff
@@ -329,8 +329,11 @@ export default class RelatedResourcesAddComponent extends Component<RelatedResou
     const relatedHermesDocument = {
       googleFileID: attrs.objectID,
       title: attrs.title,
-      type: attrs.docType,
```
jeffdaley, line 332: We were incorrectly trying to save this attribute as `type` when the backend expects `documentType`

jfreda (approved):

PR #376, opened 2023-10-20
Add `eslintrc.js` 

jeffdaley:
Adds ESLint configuration so we can add `yarn lint` to our GitHub Actions.

I've disabled the rules we're currently failing. We can enable them later, individually, as we get them to pass.


web/tests/integration/components/related-resources/add-test.ts
```diff
@@ -301,4 +301,4 @@ module("Integration | Component | related-resources/add", function (hooks) {
       .dom("[data-test-related-resources-search-loading-icon]")
       .exists("a loading icon is shown when `searchIsRunning` is true");
   });
-});
+});
```
jfreda, line 304: 😮 Was this accidental or does ESLint have something against EOF in files?

jfreda (approved):

web/tests/integration/components/related-resources/add-test.ts
jeffdaley: Accidental. In trying to quickly revert a change, I copied and pasted from GitHub, which apparently cuts off the whitespace.

PR #377, opened 2023-10-20
Add API v2

jfreda:
This PR adds a new v2 of the API, which can be enabled via feature flag. This version of the API uses PostgreSQL data as the source of truth instead of Algolia. One other change of note is that the data consistency checks for Algolia/PostgreSQL now handle slices with elements in differing order.

#### New feature flag `api_v2`:
```hcl
feature_flags {
  // api_v2 enables v2 of the API.
  flag "api_v2" {
    enabled = true
  }
}
```

### DIFF
--- a/web/app/components/document/sidebar.ts
+++ b/web/app/components/document/sidebar.ts
@@ -800,13 +800,10 @@ export default class DocumentSidebarComponent extends Component<DocumentSidebarC
    */
   private getDraftPermissions = task(async () => {
     try {
-      let apiVersion = "v1";
-      if (this.configSvc.config.feature_flags["api_v2"]) {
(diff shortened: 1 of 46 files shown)


web/app/components/document/sidebar.ts
```diff
@@ -788,8 +800,13 @@ export default class DocumentSidebarComponent extends Component<DocumentSidebarC
    */
   private getDraftPermissions = task(async () => {
     try {
+      let apiVersion = "v1";
```
jfreda, line 803: Remove this and only use config service

anubhavmishra (approved):
🎉

anubhavmishra (approved):

PR #380, opened 2023-10-20
Update "Create New" forms

jeffdaley:
- Updates the "New" forms so that "Create Doc" and "Create Project" share a layout
- Cleans the design of unnecessary labels and the document "preview"
- Tweaks error handling so that the grayed-out button provides feedback on click
- Adds an animation to the transition when the "create" task begins

https://github.com/hashicorp-forge/hermes/assets/754957/47858f86-7974-4f2a-a6e9-d247d9ec530d


web/app/components/inputs/product-select.hbs
```diff
@@ -36,11 +36,11 @@
         @offset={{@offset}}
         @matchAnchorWidth={{@matchAnchorWidth}}
         @secondaryFilterAttribute="abbreviation"
-        class="product-select-dropdown-list w-[300px]"
+        class="product-select-dropdown-list"
         ...attributes
       >
         <:anchor as |dd|>
-          <dd.ToggleSelect class="gap-2">
+          <dd.ToggleSelect class="gap-2" id="product-select">
```
jeffdaley, line 43: This is the `@controlID` value on the `label` (https://github.com/hashicorp-forge/hermes/pull/380/files#diff-ef75bd0581dfe8416249c82f232d502485f659443563d9a553c39a100ef453dbR51)

You can now click the label to open the dropdown.

web/app/components/new/doc-form.ts
```diff
@@ -13,21 +13,13 @@ import FlashService from "ember-cli-flash/services/flash-messages";
...
   productAbbreviation: string | null;
-  contributors: string | null;
 }
 
-const FORM_ERRORS: DocFormErrors = {
-  title: null,
-  summary: null,
-  productAbbreviation: null,
-  contributors: null,
-};
```
jeffdaley, line 29: This error logic was broken, but it didn't matter because the disabled state of the button prevented it from being used.

web/app/components/new/doc-form.ts
```diff
@@ -68,13 +56,10 @@ export default class NewDocFormComponent extends Component<NewDocFormComponentSi
...
-
-  /**
-   * Whether the summary is more than 200 characters.
-   * Used in the template to gently discourage long summaries.
-   */
-  @tracked protected summaryIsLong = false;
+  @tracked protected formErrors: DocFormErrors = {
+    title: null,
+    productAbbreviation: null,
+  };
```
jeffdaley, line 62: The initial implementation had the right idea of initially assigning an empty object and then validating against it. However, you'll see below that it was never updated.

web/app/components/new/doc-form.ts
```diff
@@ -116,7 +99,12 @@ export default class NewDocFormComponent extends Component<NewDocFormComponentSi
    * Validates the form and updates the `formErrors` property.
    */
   private validate() {
-    this.formErrors = { ...FORM_ERRORS };
```
jeffdaley, line 119: You can see here that we were repeatedly assigning an empty object to the `formErrors` property instead of having it reflect the actual state of the form.

web/app/components/new/doc-form.ts
```diff
@@ -148,10 +135,17 @@ export default class NewDocFormComponent extends Component<NewDocFormComponentSi
       this.productArea = formObject["productArea"] as string;
     }
 
-    if (this.summary.length > 200) {
-      this.summaryIsLong = true;
-    } else {
-      this.summaryIsLong = false;
```
jeffdaley, line 154: I removed this because it no longer seemed necessary.

We initially implemented it to reduce the likelihood of [unclean string] errors (i.e., strings with special characters). Yes, we also wanted succinct summaries, but that wasn't the primary driver, nor was it the problem.

Any thoughts?

web/app/components/new/doc-form.ts
```diff
@@ -148,10 +135,17 @@ export default class NewDocFormComponent extends Component<NewDocFormComponentSi
...
+    if (e.key === "Enter") {
+      e.preventDefault();
```
jeffdaley, line 139: Our current implementation allows newlines in the summary field 🙊

web/app/components/new/doc-form.ts
jfreda: Hmm, I think this was because we saw issues with the doc header rewrites if there was a very long summary - I haven't tested that since we added this restriction in the frontend though. Can we add this back for now? The backend should be enforcing some length maximum but isn't currently (then we can maybe remove this?).

jfreda (approved):
One comment about keeping the summary length maximum, but everything else looks 👍

web/app/components/new/doc-form.ts
jeffdaley: Added it back, but to be clear, this isn't a restriction,but a suggestion (though it'd take a real REBEL to disregard the warning text).

PR #385, opened 2023-10-26
Expand HDS template types

jeffdaley:
- <strike>Adds the `number` type to the `text` argument of the `Hds::BadgeCount` and `Hds::Badge` Glint types. The docs specify `string` but in reality they both take `any` value; I'm specifying `string | number` as an internal type check.</strike>
- Makes the `FlightIcon` component available in `gts` files.

### DIFF
--- a/web/types/hds/badge-count.d.ts
+++ b/web/types/hds/badge-count.d.ts
@@ -1,4 +1,4 @@
-//helios.hashicorp.design/components/badge-count?tab=code#component-api
+// helios.hashicorp.design/components/badge-count?tab=code#component-api
 
 import { ComponentLike } from "@glint/template";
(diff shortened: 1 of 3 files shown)


jfreda:
> * Adds the `number` type to the `text` argument of the `Hds::BadgeCount` and `Hds::Badge` Glint types. The docs specify `string` but in reality they both take `any` value; I'm specifying `string | number` as an internal type check.

Hmm, wondering how risky it is to diverge from the docs here. It seems reasonable that the `any` could change to `string` in the underlying implementation without notice if the docs are considered the API contract. Do we actually need to use `number` for this?

jfreda (approved):

PR #386, opened 2023-10-26
Add `JiraObject` type

jeffdaley:
Adds `JiraObject` type to the `Project` type declaration file.

### DIFF
--- a/web/app/types/project.d.ts
+++ b/web/app/types/project.d.ts
@@ -2,22 +2,26 @@ import {
   RelatedExternalLink,
   RelatedHermesDocument,
 } from "hermes/components/related-resources";
+import { ProjectStatus } from "./project-status";
+
+export interface JiraObject {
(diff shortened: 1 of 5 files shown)


web/app/types/project.d.ts
```diff
@@ -2,22 +2,26 @@ import {
...
 } from "hermes/components/related-resources";
+import { ProjectStatus } from "./project-status";
+
+export interface JiraObject {
+  key: string;
+  url: string;
+  priority: string;
+  status: string;
+  assignee?: string;
+  type?: string;
```
jfreda, line 13: Is there a case where Jira objects won't have a type?

jfreda (approved):

web/app/types/project.d.ts
jeffdaley: Yeah probably not. Updated.

PR #394, opened 2023-10-27
Create `Product::Avatar` component

jeffdaley:
Pending #385 #387 

Component-izes the "product badge" (renamed to ProductAvatar). We'll soon expand this to handle products without FlightIcons.


web/app/components/doc/thumbnail.ts
```diff
@@ -20,6 +25,14 @@ export default class DocThumbnailComponent extends Component<DocThumbnailCompone
     }
   }
 
+  protected get productAvatarSize() {
```
jfreda, line 28: Do we still need `sizeIsLarge()` with this?

web/app/components/product/avatar.gts
```diff
@@ -0,0 +1,59 @@
...
+    return productID;
+  }
+
+  private get sizeIsMedium() {
+    return this.args.size === HermesAvatarSize.Medium;
+  }
+
+  private get sizeIsLarge() {
+    return this.args.size === HermesAvatarSize.Large;
+  }
```
jfreda, line 35: Curious about this pattern vs. checking the value in the template using a more generic getter that just returns the size (like was added in the thumbnail component in this PR) - seems like that could be more scalable?

jfreda (commented):
Overall looks good, just couple of questions about the pattern for preferring boolean getters so I can understand better.

web/app/components/product/avatar.gts
jeffdaley: Most of it was the convenience of having everything in one file in the early phase. But it's worth noting that Tailwind's [dynamic class rules](https://tailwindcss.com/docs/content-configuration#dynamic-class-names) limit our efficiency a bit.* Still, we're past the early phase, so we can be smarter here. I refactored to use CSS classes.

* Of course, there are workarounds. I think safe-listing the `w-` and `h-` classes would allow us to do stuff like `class="h-{{this.size}}"` if that were ever preferable to CSS.

jfreda: Cool, sounds good 👍

jfreda (approved):

PR #397, opened 2023-10-27
Add `large` and `xl` sizes to PersonAvatar

jeffdaley:
Adds `large` and `xl` sizes to and writes tests for the `Person::Avatar` component.


web/app/types/avatar-size.ts
```diff
@@ -1,5 +1,13 @@
-export enum HermesAvatarSize {
+export enum HermesBasicAvatarSize {
   Small = "small",
   Medium = "medium",
   Large = "large",
 }
+
+export enum HermesExtendedAvatarSize {
+  XL = "xl",
+}
```
jfreda, line 9: Curious why we need both of these instead of one for all avatar sizes?

jfreda (approved):

web/app/types/avatar-size.ts
jeffdaley: Currently only the `BasicSize`s are shared between Product and Person avatars. So we type `ProductAvatar` as `BasicSize` and `PersonAvatar` as `BasicSize | ExtendedSize`. Main goal is to make Glint squiggle if you try to pass `xl` to `ProductAvatar`.

PR #406, opened 2023-11-06
Fetch products on login; add `getProductAbbreviation` function

jeffdaley:
Moves the ProductAreaService's `fetch` call to the `/authenticated` route and adds a `getProductAbbreviation` utility method. We now fetch ProductAreas at the component level (e.g., ProductSelect, Settings), but since we're planning to use abbreviations in the ProductAvatar, it would be helpful to have these loaded on login.


jfreda (commented):
If I'm understanding this correctly, we'll only fetch the products on login with this change so if (in the probably rare case, but possible) a product is added while a user has an active session, it could cause an issue? If that's correct, should we do another fetch if we don't find a product in the productArea service?

jeffdaley:
> if a product is added while a user has an active session, it could cause an issue?

FWIW, this wouldn't be a new issue. Even in the current code, ProductAreas are only fetched once per session.

I added a couple "refresh" calls to handle the case of new products being added. I think they make sense on the [ProductSelect](https://github.com/hashicorp-forge/hermes/pull/406/commits/56bc2fd2a44b931b79fcbb509b314dce214e7d5b) and [SubscriptionList](https://github.com/hashicorp-forge/hermes/pull/406/commits/50aa9979fdb00b566dec8f6e1c73b79732986302) components but don't seem necessary for cosmetic stuff like avatars.

jfreda:
> FWIW, this wouldn't be a new issue.

Ah didn't realize that.

> I added a couple "refresh" calls to handle the case of new products being added.

Cool, sounds good 👍

jfreda (approved):

PR #408, opened 2023-11-07
Update request for project related resources; Cleanup to match back-end

jeffdaley:
- Updates the Project requests to grab the related resources separately.
- Renames stuff to align with the back end
- Cleanup and note-to-self


web/mirage/config.ts
```diff
@@ -516,24 +528,6 @@ export default function (mirageConfig) {
...
-      /**
-       * Used by the /projects/:project_id route to fetch a single project.
-       */
-      this.get("/projects/:project_id", (schema, request) => {
-        const project = schema.projects.findBy({
-          id: request.params.project_id,
-        });
-        return new Response(200, {}, project.attrs);
-      });
-
```
jeffdaley, line 536: These were duplicates. Probably a merge error.

web/app/types/project.d.ts
```diff
@@ -15,14 +15,15 @@ export interface JiraObject {
...
 export interface HermesProject {
-  id: string; // at least in Mirage...
+  id: string;
   title: string;
   status: ProjectStatus;
   hermesDocuments?: RelatedHermesDocument[];
   description?: string;
-  jiraObject?: JiraObject;
+  jiraIssueID?: string;
+  jiraIssue?: JiraIssue;
```
jfreda, line 24: Is `jiraIssueID` supposed to represent the internal Jira ID (if so, curious how we use that)? Can we put that as part of a `JiraIssue`? Note/caveat: I haven't dove into the Jira part of the backend work yet so I'm cool with this changing as we start to incorporate that.

jfreda (approved):

web/app/types/project.d.ts
jeffdaley: I interpreted JiraIssueID to be an _external_ reference, like "LABS-123" (as in `test-org.atlassian.com/issues/LABS-123` that would be used to query the Jira API for the full object.

PR #414, opened 2023-11-08
Projects in the sidebar

jeffdaley:
Adds a Projects section to the document sidebar along with "add document to project" functionality.


web/app/components/document/sidebar.hbs
```diff
@@ -11,15 +11,15 @@
 
 {{#unless @isCollapsed}}
   <div
-    class="sidebar-body scrollable-container"
```
jeffdaley, line 14: Unused class

web/app/components/document/sidebar.ts
```diff
@@ -477,6 +507,23 @@ export default class DocumentSidebarComponent extends Component<DocumentSidebarC
...
+
+  /**
+   * The action to hide the Projects modal.
+   * Passed to the Projects modal component and
+   * triggered on modal close.
+   */
+  @action protected hideProjectsModal() {
+    this.projectsModalIsShown = false;
+  }
+
```
jfreda, line 526: Non-blocking: maybe could simplify to a `toggleProjectsModall()` that does `this.projectsModalIsShown = !this.projectsModalIsShown`?

web/app/components/new/project-form.ts
```diff
@@ -107,6 +122,31 @@ export default class NewProjectFormComponent extends Component<NewProjectFormCom
           }),
         })
         .then((response) => response?.json());
+
+      /**
+       * Create the project with a minimum duration.
```
jfreda, line 127: 2 fast 2 furious

web/app/components/project/jira-widget.ts
```diff
@@ -75,7 +75,72 @@ export default class ProjectJiraWidgetComponent extends Component<ProjectJiraWid
...
+      return this.issue.status as string;
+    }
+  }
+
+  /**
+   * The type of the issue, if it exists.
+   * Will exist if the issue is a full Jira issue.
+   * Determines if the type is shown in the widget.
+   */
+  protected get issueType() {
```
jfreda, line 87: Curious why we need these getters and can't just do conditionals in the template based on if the issue's properties are undefined or not?

web/app/components/projects/add-to-or-create.hbs
```diff
@@ -0,0 +1,100 @@
+<Hds::Modal
+  data-test-add-to-or-create-project-modal
+  @onClose={{@onClose}}
+  class={{unless this.newProjectFormIsShown "search-window"}}
+  as |M|
+>
+  <M.Header data-test-modal-header>
+    {{if this.newProjectFormIsShown "Start a project" "Add doc to project"}}
+  </M.Header>
+  {{! TODO: confirm this height decision }}
```
jfreda, line 10: Did you mean to leave this TODO in? (fine if it will be addressed later, but wanted to make sure)

jfreda (approved):
Couple questions, but nothing blocking!

web/app/components/document/sidebar.ts
jeffdaley: I usually reserve the `toggle` construction for actual toggles, e.g., "Show more/Show fewer" or check/uncheck controls. Here, because the show and hide actions are separate, I like the clarity of being specific: <button {{on "click" this.showProjectsModal}}>` and `<Hds::Modal @onClose={{this.hideProjectsModal}}>` feels more precise to me than if we were to pass `toggleProjectsModal` to both.

Also, it's not the case here, but show/hide actions often have specific setup or teardown, so I might also be in the habit of separating them from the jump.

web/app/components/project/jira-widget.ts
jeffdaley: For some reason Glint has trouble when using union types like this

```
{{! this.issue: PickedJiraIssue | JiraIssue }}

{{#if this.issue.issueType}}
   {{!-- if this passes, it means this.issue is type: JiraIssue --}}
   {{!-- however Glint thinks it still might be a PickedJiraIssue --}}
   
   {{this.issue.issueType}}
   {{!-- this gets underlined by Glint --}}
{{/if}}
```

web/app/components/projects/add-to-or-create.hbs
jeffdaley: nope, good catch

PR #415, opened 2023-11-08
Allow approving documents in Approved status

jfreda:
'Nuff said.

### DIFF
--- a/internal/api/approvals.go
+++ b/internal/api/approvals.go
@@ -354,9 +354,9 @@ func ApprovalHandler(
 
 			// Authorize request.
 			userEmail := r.Context().Value("userEmail").(string)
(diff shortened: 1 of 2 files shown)


internal/api/approvals.go
```diff
@@ -354,9 +354,9 @@ func ApprovalHandler(
 
 			// Authorize request.
 			userEmail := r.Context().Value("userEmail").(string)
-			if doc.Status != "In-Review" && doc.Status != "In Review" {
```
jfreda, line 357: We never actually needed to confirm the "In Review" status because all docs created through Hermes would only have an "In-Review" (with the hyphen) status.

anubhavmishra (approved):

PR #420, opened 2023-11-15
Combine My Docs and My Drafts

jeffdaley:
My Docs

CleanShot 2023-11-22 at 13 23 59@2x

- Combines the `/drafts` and `/my` routes into a view at `/my/documents`
- Replaces facet filters with a single toggle to show/hide drafts you're collaborating on.
- Groups documents by modified date, e.g., "Recently active" and "More than 30 days old" 
- Increases the number of docs per page to minimize the need for extra controls

**Why remove filtering?**

Primary to simplify the UX. Most users don't yet have enough docs to make filtering necessary. The use case isn't there the same way it is for All Docs, where there's thousands of docs spread across dozens of products. We're also under the current limitation of Drafts and Published Docs being on separate indexes, which makes filtering (let alone pagination) a high-cost, low-value task. Long-term, I expect we'll add back some lightweight filters, but it'll be once Drafts and Docs are a single index.


jfreda (approved):
Code looks good - we should discuss a potential move away from Algolia for drafts in general if we're not going to support searching them, and how the Hermes API could enable this for the frontend.

PR #423, opened 2023-11-15
Add `timeColumn` argument to `Table::Row`

jeffdaley:
Adds a `timeColumn` argument to the `Table::Row` component to control whether the `createdTime` or `modifiedTime` is displayed.


web/app/components/table/row.ts
```diff
@@ -1,31 +1,40 @@
...
 
-    if (!docOwner) {
-      return false;
+    if (modifiedTime && timeColumn === TimeColumn.Modified) {
+      time = modifiedTime;
     }
 
-    return docOwner === this.authenticatedUser.info.email;
-  }
+    if (createdTime && timeColumn === TimeColumn.Created) {
```
jfreda, line 29: Super-nit: could probably be an `else if` (for a HUGE performance boost /s)

PR #426, opened 2023-11-15
Standardize `FlashMessage` properties

jeffdaley:
Adds default configurations for the FlashMessage service as well as `critical()` and `success()` helper methods. Standardizes the duration of long timeouts. We'll now import from an extended service (`HermesFlashMessagesService`) which includes types for the new utility methods.

### DIFF
--- a/web/config/environment.js
+++ b/web/config/environment.js
@@ -5,7 +5,7 @@ const getEnv = (key, defaultValue) => {
   const value = process.env[fullKey];
   if (value == null) {
     console.warn(
-      `env var ${fullKey} was not set! Proceeding with default value "${defaultValue}"`
+      `env var ${fullKey} was not set! Proceeding with default value "${defaultValue}"`,
     );
   }
   return value != null ? value : defaultValue;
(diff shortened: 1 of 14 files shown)


web/config/environment.js
```diff
@@ -49,6 +49,13 @@ module.exports = function (environment) {
       apiKey: getEnv("ALGOLIA_SEARCH_API_KEY"),
     },
 
+    flashMessageDefaults: {
+      timeout: 5000,
+      extendedTimeout: 1000,
+      type: "success",
+      types: ["critical", "success"],
+    },
```
jfreda, line 57: What's the use case(s) you're thinking for making these configurable per environment?

jeffdaley: No plans for environment-specific config yet (though I could see using short timeouts if `environment === "testing"`). I'm just using `environment.js` to be in accordance with the [addon configuration](https://github.com/adopted-ember-addons/ember-cli-flash#service-defaults).

jfreda (approved):

PR #432, opened 2023-11-16
Move Dashboard from template to component

jeffdaley:
Creates a `Dashboard` component out of the current template and controller. This would generally be the first step in any Dashboard changes, so I wanted to isolate it to make future diffs easier.


jfreda (commented):
Couple of high-level questions:
- Why are component-izing the dashboard? Is this component going to be reused somewhere else?
- Any reason we're not using the single-file component style? (I thought we were moving to that?)

jeffdaley:
1. We've component-ized almost every template by now. Working with co-located files is more comfortable than jumping between folders. It also keeps routes and controllers focused only on redirecting, queryParams and models, with components doing everything else.
2. Unfortunately `gts` is a little rough with our version of Ember (or at least my setup). I'm not getting import suggestions which makes writing stuff like `{{#if (or this.isSelected (eq this.count 3))}}` more difficult than it currently is. I'm still gonna try to use it, but it seems safer on lower-level components than Big Daddies.

jfreda:
> * We've component-ized almost every template by now. Working with co-located files is more comfortable than jumping between folders. It also keeps routes and controllers focused only on redirecting, queryParams and models, with components doing everything else.
> * Unfortunately `gts` is a little rough with our version of Ember (or at least my setup). I'm not getting import suggestions which makes writing stuff like `{{#if (or this.isSelected (eq this.count 3))}}` more difficult than it currently is. I'm still gonna try to use it, but it seems safer on lower-level components than Big Daddies.

Okay works for me, thanks 👍

jfreda (approved):

PR #435, opened 2023-11-17
Bugfix: Add `?draft=true` to draft links

jeffdaley:
Fixes (and now tests for) the `draft` queryParam on `/draft` document links. 

Bug was introduced #418 by not capitalizing "WIP."

### DIFF
--- a/web/app/components/table/row.ts
+++ b/web/app/components/table/row.ts
@@ -13,6 +13,14 @@ interface TableRowComponentSignature {
 export default class TableRowComponent extends Component<TableRowComponentSignature> {
   @service declare authenticatedUser: AuthenticatedUserService;
 
+  protected get isDraft() {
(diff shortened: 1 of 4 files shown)


web/app/components/table/row.ts
```diff
@@ -13,6 +13,14 @@ interface TableRowComponentSignature {
 export default class TableRowComponent extends Component<TableRowComponentSignature> {
   @service declare authenticatedUser: AuthenticatedUserService;
 
+  protected get isDraft() {
+    if (this.args.doc.isDraft) {
+      return true;
+    }
+
+    return this.args.doc.status === "WIP";
```
jeffdaley, line 21: I'll `enum` these statuses separately.

jfreda (approved):

PR #438, opened 2023-11-20
PENDING: Dashboard improvements

jeffdaley:
Reworks the dashboard to be more performant, information-dense and nice-looking. Lays the foundation for a "Following"-style feed and will make it easier to include projects in Recently Viewed.

CleanShot 2023-11-20 at 20 54 43@2x

- Increases the number of recently viewed docs from 4 → 10
- Turns the "latest updates" widget into a feed-like design that shows the entire first page of published docs
- Moves the "latest docs" concept into a service where we can be smarter about fetching and caching. Similar to `RecentlyViewedDocs`, we now only `await` the `fetchAll` promise if the `index === null`. In other cases we still check for new docs, but we don't delay rendering the template. (Yes, this can result in pop-in, but it's rare enough and minimally detrimental that it shouldn't hold up merging.)
- Adds tests to a route we've been neglecting


jeffdaley:
To be recast in a separate PR

PR #442, opened 2023-11-21
Fix project-flag routing bug

jeffdaley:
Adds projects routes to the Router, along with redirects to the dashboard if the feature flag isn't enabled.


web/app/routes/authenticated/new/project.ts
```diff
@@ -1,3 +1,15 @@
...
+import RouterService from "@ember/routing/router-service";
+import { inject as service } from "@ember/service";
+import FlagsService from "hermes/services/flags";
 
-export default class AuthenticatedNewProjectRoute extends Route {}
+export default class AuthenticatedNewProjectRoute extends Route {
+  @service declare flags: FlagsService;
+  @service declare router: RouterService;
+
+  beforeModel() {
```
jfreda, line 10: It looks like the `redirect()` method might be preferred for this use case? https://guides.emberjs.com/v3.28.0/routing/redirection/#toc_child-routes

web/app/routes/authenticated/projects.ts
```diff
@@ -1,3 +1,15 @@
...
+import RouterService from "@ember/routing/router-service";
+import { inject as service } from "@ember/service";
+import FlagsService from "hermes/services/flags";
 
-export default class AuthenticatedProjectsRoute extends Route {}
+export default class AuthenticatedProjectsRoute extends Route {
+  @service declare flags: FlagsService;
+  @service declare router: RouterService;
+
+  beforeModel() {
```
jfreda, line 10: Same as above - `redirect()` instead because this is also a child route?

web/app/routes/authenticated/new/project.ts
jeffdaley: I think that recommendation applies to what would otherwise be `afterModel` redirects. (That section is nested under "Transitioning after the model is known" which isn't the case for us.) I'm going off the section above that https://guides.emberjs.com/release/routing/redirection/#toc_transitioning-before-the-model-is-known

jfreda: Ah, got it, makes sense. I was trying to figure out if that was the case from the docs and couldn't tell.

jfreda (approved):

PR #454, opened 2023-11-28
Switch `patchDocument` from `task` to `enqueueTask`

jeffdaley:
Makes the `patchDocument` task enqueue-able to handle instances where two patches fire at once, corrupting the document. If an `EditableField` is open and has changes, clicking the "change status" button (e.g., "Move to approved") will trigger two patch calls: one to save the EditableField change; the other to save the status. Now, Hermes waits for the first task to finish before starting the second.


jfreda (approved):
Nice catch!

PR #456, opened 2023-11-28
Improve locked doc state

jeffdaley:
Reduces confusion of the locked-doc state by improving the precision of the `isReadOnly` logic in the sidebar. We were incorrectly allowing some fields to be (or at least appear) editable by a document owner even if the doc was in a locked state. We now check for this.


web/app/components/new/doc-form.ts
```diff
@@ -209,6 +209,7 @@ export default class NewDocFormComponent extends Component<NewDocFormComponentSi
             productAbbreviation: this.productAbbreviation,
             summary: cleanString(this.summary),
             title: cleanString(this.title),
+            appCreated: true,
```
jfreda, line 212: This is automatically set by the backend and not needed.

jfreda (approved):
One change suggested, otherwise looks good

PR #459, opened 2023-11-28
Component-ize Project documents (`Doc::TileMedium`)

jeffdaley:
Converts project-document code into a `Doc::TileMedium` component which can be used on other screens. Includes logic to load the ownerPhotos of newly added docs.

(I imagine we'll eventually combine the `Doc::Tile` components but I'd like to handle that separately.)


web/app/styles/components/project.scss
```diff
@@ -71,7 +71,7 @@
     }
 
     .overflow-button-container {
-      @apply absolute right-1 top-1;
+      @apply absolute right-1 top-1 z-10 h-[calc(100%-16px)] items-start;
```
jeffdaley, line 74: The overflow-button gradient was still slightly visible even after #452

web/tests/acceptance/authenticated/projects/project-test.ts
```diff
@@ -217,15 +217,9 @@ module("Acceptance | authenticated/projects/project", function (hooks) {
...
-      .dom(DOCUMENT_TYPE)
-      .containsText(docType)
-      .hasAttribute("href", "/documents?docType=%5B%22PRD%22%5D");
+    assert.dom(DOCUMENT_TYPE).containsText(docType);
 
-    assert
-      .dom(DOCUMENT_STATUS)
-      .containsText(docStatus)
-      .hasAttribute("href", "/documents?status=%5B%22Approved%22%5D");
+    assert.dom(DOCUMENT_STATUS).containsText(docStatus);
```
jeffdaley, line 222: I removed interactivity from the status and docType tags. The use case for these filters isn't very strong, especially in comparison to docOwner/productArea — themselves debatable as links in this context — so as not to be worth sacrificing the click area.

jfreda (approved):

PR #469, opened 2023-12-01
Improve approving-your-own-doc UX

jeffdaley:
Improves the UX of approving your own doc. We now hit the `approvals` endpoint as part of the "Move to approved" flow which adds a checkmark to the owner-approver in Hermes as well as the document.


jfreda (approved):
I'd prefer that we don't allow adding a doc owner as an approver of their own doc, but approving because that's not the case right now.

PR #470, opened 2023-12-01
Add `status` as a project `AttributeForFaceting`

jeffdaley:
Configures the projects index to allow filtering by `status`.


jeffdaley:
Closing since we want to move away from Algolia for non-global searches. In the short term we'll do local status filtering.

PR #475, opened 2023-12-05
Add status nav to Projects index

jeffdaley:
Adds a status nav to the projects index. For now we fetch all results and filter locally but we'll soon add support for back-end filtering. I'm keeping this mostly undesigned for now, so as not to compete with any open PRs.


web/app/controllers/authenticated/projects/index.ts
```diff
@@ -1,7 +1,12 @@
 import Controller from "@ember/controller";
 import AuthenticatedProjectsIndexRoute from "hermes/routes/authenticated/projects/index";
+import { ProjectStatus } from "hermes/types/project-status";
 import { ModelFrom } from "hermes/types/route-models";
 
 export default class AuthenticatedProjectsController extends Controller {
+  queryParams = ["status"];
+
+  status = ProjectStatus.Active;
```
jfreda, line 9: What is this line doing?

jfreda (approved):
One question, otherwise looks good

web/app/controllers/authenticated/projects/index.ts
jfreda: Ah got it, maybe a comment would help clarify for anyone else besides me who didn't immediately understand this 🙂

PR #480, opened 2023-12-05
Add document type to Google Doc header title

jfreda:
This PR adds the document type to the (Google Doc) header title so it now has the format of `[RFC] ABC-123: Title of my doc`. Without this, the document type isn't explicitly in the Google Doc so it can make things confusing.

This PR also fixes a small bug when checking equality of Algolia vs. database doc summary where we were logging the memory address of the database summary instead of the text (missed this when converting the summary to a pointer).


anubhavmishra (approved):

anubhavmishra:
Easy change!

PR #502, opened 2023-12-14
Check for `dialog`s when using FloatingUI + `renderOut`

jeffdaley:
Adds a check for `dialog` in Floating UI's `renderOut` logic. Normally we render out to the bottom of the DOM and use `z-index` for layering, but because `dialog` is  part of the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) we need to render into it to overlay it.

### DIFF
--- a/web/app/components/floating-u-i/content.ts
+++ b/web/app/components/floating-u-i/content.ts
@@ -52,7 +52,7 @@ export default class FloatingUIContent extends Component<FloatingUIContentSignat
    */
   protected get inElementTarget() {
     const dialog = document.querySelector("dialog");
-    console.log("dialog", dialog);
+
     if (dialog) {
       return "dialog";
     } else {


web/app/components/floating-u-i/content.ts
```diff
@@ -45,6 +45,21 @@ export default class FloatingUIContent extends Component<FloatingUIContentSignat
...
   }
 
+  /**
+   * The target element to append the content to.
+   * If a dialog is open, append to it, since it's the
+   * topmost layer, otherwise append to the application body.
+   */
+  protected get inElementTarget() {
+    const dialog = document.querySelector("dialog");
+    console.log("dialog", dialog);
```
jfreda, line 55: Did you mean to keep this logging in?

jfreda (approved):

web/app/components/floating-u-i/content.ts
jeffdaley: Thanks for catching

PR #504, opened 2023-12-14
Remove partial comment and a couple `console.logs`

jeffdaley:


web/mirage/config.ts
```diff
@@ -148,7 +148,6 @@ export default function (mirageConfig) {
                   return doc.attrs.product === product;
                 });
               } else {
-                // This
```
jeffdaley, line 151: changes everything

jfreda (approved):

PR #508, opened 2023-12-15
Enable creating documents as logged-in user

jfreda:
This PR enables creating documents on behalf of logged-in Hermes users when running Hermes as a service account with domain-wide authority.

### New config attribute

```hcl
google_workspace {
  ...
  auth {
    ...
    create_docs_as_user = true
    ...
  }
  ...
}
```

### DIFF
--- a/configs/config.hcl
+++ b/configs/config.hcl
@@ -131,10 +131,11 @@ google_workspace {
   // auth is the configuration for interacting with Google Workspace using a
   // service account.
   // auth {
-  //   client_email = ""
-  //   private_key  = ""
-  //   subject      = ""
-  //   token_url    = "https://oauth2.googleapis.com/token"
+  //   client_email        = ""
+  //   create_docs_as_user = true
+  //   private_key         = ""
+  //   subject             = ""
+  //   token_url           = "https://oauth2.googleapis.com/token"
   // }
 
   // oauth2 is the configuration used to authenticate users via Google.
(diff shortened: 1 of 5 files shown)


configs/config.hcl
```diff
@@ -131,10 +131,11 @@ google_workspace {
   // auth is the configuration for interacting with Google Workspace using a
   // service account.
   // auth {
-  //   client_email = ""
-  //   private_key  = ""
-  //   subject      = ""
-  //   token_url    = "https://oauth2.googleapis.com/token"
+  //   client_email        = ""
+  //   create_docs_as_user = true
```
anubhavmishra, line 135: that's one way to make it easy for our users! Great work @jfreda 🎉

web/web.go
```diff
@@ -121,6 +129,7 @@ func ConfigHandler(
 			AlgoliaDocsIndexName:     cfg.Algolia.DocsIndexName,
 			AlgoliaDraftsIndexName:   cfg.Algolia.DraftsIndexName,
 			AlgoliaInternalIndexName: cfg.Algolia.InternalIndexName,
+			CreateDocsAsUser:         createDocsAsUser,
```
anubhavmishra, line 132: Curious why we need to send it in the config?

anubhavmishra (approved):

web/web.go
jfreda: I was thinking that we could make the modal that pops up after a draft is created (or at least the text in it about users having to manually manage their notifications) conditional on this.

anubhavmishra: That is a good idea

PR #519, opened 2023-12-20
Add `isSaving` state to project title and description

jeffdaley:
Adds an `isSaving` spinner to the project title and description fields.

### DIFF
--- a/web/app/components/project/index.hbs
+++ b/web/app/components/project/index.hbs
@@ -79,6 +79,7 @@
       @placeholder="Add project title"
       @onSave={{this.saveTitle}}
       @isRequired={{true}}
+      @isSaving={{this.titleIsSaving}}
     />
   </div>
 
@@ -90,6 +91,7 @@
       @value="{{this.description}}"
       @onSave={{this.saveDescription}}
       @placeholder="Add a description"
+      @isSaving={{this.descriptionIsSaving}}
     />
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


jfreda (commented):
If I'm understanding this correctly, I'm not sure it would work if someone were to very quickly save a title and then a description because this is tracking each saving state separately and we're not doing any enqueueing of the async tasks (and the save function as-is wouldn't be able to differentiate between saving each field). It seems like the easiest thing would be to just make one saving state (like we do for the sidebar), or else make changes to support saving each field separately. Let me know if you think I'm misunderstanding the logic though.

jeffdaley:
> If I'm understanding this correctly, I'm not sure it would work if someone were to very quickly save a title and then a description because this is tracking each saving state separately and we're not doing any enqueueing of the async tasks (and the save function as-is wouldn't be able to differentiate between saving each field). It seems like the easiest thing would be to just make one saving state (like we do for the sidebar), or else make changes to support saving each field separately. Let me know if you think I'm misunderstanding the logic though.

I changed the `save` task to be `restartable` to handle the edge case where a save attempt is made while another is in progress. In this case, you'd see two spinners. For general usage (editing one at a time, respecting the short `isSaving` state), a single spinner looks better.

jfreda:
> I changed the `save` task to be `restartable` to handle the edge case where a save attempt is made while another is in progress. In this case, you'd see two spinners. For general usage (editing one at a time, respecting the short `isSaving` state), a single spinner looks better.

Okay, yeah agree about the UX, but I was just thinking about the functionality. I'm wondering if there still might be an edge case where the first task fails and the request gets lost, but I'm okay with leaving that for later if it's a possibility.

jfreda (approved):

jeffdaley:
Ah, looking back at it, I was under the assumption that we always saved title, description and status regardless of which attribute was changing. That isn't the case so I'm switching to an `enqueue` task. Sorry for confusing things.

PR #528, opened 2023-12-21
Add read-only styles to Jira

jeffdaley:
Adds read-only styles to the Jira widget. Inactive projects will no longer have the Jira overflow button (to remove). For inactive projects without a Jira issue, the widget is removed entirely.

Also updates the Jira widget to use the same OverflowMenu settings as the project resource tiles. Now when the Jira issue is at max width, its overflow button aligns vertically with the resource overflow buttons 🥹

### DIFF
--- a/web/app/components/project/jira-widget.hbs
+++ b/web/app/components/project/jira-widget.hbs
@@ -1,232 +1,213 @@
-<div
-  class="jira-widget relative flex h-[36px] w-full max-w-3xl items-center"
-  ...attributes
->
+{{#unless (and (not this.issue) (not @isLoading) @isReadOnly)}}
   <div
-    class="inner relative flex w-full
-      {{if
-        (and this.issue @isNewProjectForm)
-        'h-9 items-center rounded-button-md pr-2 shadow-surface-low'
-      }}"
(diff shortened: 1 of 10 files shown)


web/app/components/project/jira-widget.hbs
```diff
@@ -1,232 +1,213 @@
-<div
-  class="jira-widget relative flex h-[36px] w-full max-w-3xl items-center"
-  ...attributes
->
+{{#unless (and (not this.issue) (not @isLoading) @isReadOnly)}}
```
jfreda, line 1: Style: this logic might be more readable if it was converted to a getter?

jfreda (approved):

web/app/components/project/jira-widget.hbs
jeffdaley: Yeah the `unless` makes it extra confusing.

I'll wrap this change into #523, because I've already moved the check into the parent component to be used in an  `animated-if` statement. It'll soon be

```
 get jiraWidgetIsShown() {
    return (
      this.projectIsActive || this.jiraIssue || this.loadJiraIssue.isRunning
    );
  }
```

PR #529, opened 2023-12-22
Configure `createDocsAsUser` in the front end

jeffdaley:
Configures `createDocsAsUser` property to conditionally replace the "Draft Created" modal with a flash message.


jfreda (approved):
🚀

PR #531, opened 2023-12-22
Fix Jira render logic

jeffdaley:
Fixes a bug causing inactive projects' Jira issues not to load. Also improves header padding when the Jira widget is not present.

### DIFF
--- a/web/app/components/project/index.hbs
+++ b/web/app/components/project/index.hbs
@@ -64,7 +64,7 @@
 
   </div>
 </div>
-<div class="relative z-20 mb-5 max-w-3xl">
+<div class="relative z-20 max-w-3xl pb-5">
 
   {{! Title }}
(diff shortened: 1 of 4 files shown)


web/app/components/project/index.hbs
```diff
@@ -99,16 +99,19 @@
 </div>
 
 {{#if this.jiraIsEnabled}}
-  {{! Jira }}
-  <div class="mt-10 flex w-full max-w-3xl">
-    <Project::JiraWidget
-      {{did-insert this.maybeLoadJiraInfo}}
```
jeffdaley, line 105: This was not getting called because of `JiraWidget`'s previous show/hide logic. The element with `attributes` wasn't being rendered so this `did-insert` went nowhere, causing the `(not @isLoading)` check to incorrectly fail.

PR #536, opened 2024-01-08
Make sure project docs are excluded from search results

jeffdaley:
Adds logic to exclude project documents from the "add resource" results.


jfreda (approved):
Good catch

PR #553, opened 2024-01-18
Remove unused `animateHighlight` code from project component

jeffdaley:
Removes an unused argument from the `saveProjectResources` task. The highlight-element function was moved into `project/resource-list`.


jfreda (approved):
RIP

PR #558, opened 2024-01-23
Configure `Person` in EmberData

jeffdaley:
Prepares the app to use EmberData to handle People requests. This will fix avatars, add `name` attributes, and add caching.

Calls to the `people` endpoint will soon change from, e.g.,

```
await this.fetchSvc
  .fetch(`/api/${this.configSvc.config.api_version}/people`, {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify({
      query: query,
    }),
  },
);
```
to something like

```
await this.store.query("person", { query });
```

I'll follow this up with an implementation PR.

### DIFF
--- a/web/app/adapters/application.ts
+++ b/web/app/adapters/application.ts
@@ -1,0 +1,22 @@
+import JSONAdapter from "@ember-data/adapter/json-api";
+import { inject as service } from "@ember/service";
+import ConfigService from "hermes/services/config";
+import FetchService from "hermes/services/fetch";
+import SessionService from "hermes/services/session";
+
+export default class ApplicationAdapter extends JSONAdapter {
+  @service("config") declare configSvc: ConfigService;
+  @service("fetch") declare fetchSvc: FetchService;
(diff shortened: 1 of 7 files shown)


web/app/adapters/application.ts
```diff
@@ -0,0 +1,22 @@
...
+  get namespace() {
+    return `api/${this.configSvc.config.api_version}`;
+  }
+
+  get headers() {
+    return {
+      "Hermes-Google-Access-Token":
+        this.session.data.authenticated.access_token,
+    };
+  }
```
jfreda, line 21: Are these needed/used?

jfreda (approved):
Wondering if some code maybe isn't needed but otherwise LGTM 🚀

web/app/adapters/application.ts
jeffdaley: These will come into play when we start fetching with `this.store` methods.

PR #561, opened 2024-01-25
Use EmberData for People

jeffdaley:
Uses EmberData to fetch and cache People information to replace the spotty Algolia `ownerPhotos` value as the source of avatar URLs. Changes people labels from email addresses to displayNames (where present).


web/app/services/authenticated-user.ts
```diff
@@ -70,9 +63,14 @@ export default class AuthenticatedUserService extends Service {
    */
   loadInfo = task(async () => {
     try {
-      this._info = await this.fetchSvc
-        .fetch(`/api/${this.configSvc.config.api_version}/me`)
-        .then((response) => response?.json());
+      const mes = await this.store.findAll("me");
```
jfreda, line 66: All the me-s

jfreda (approved):
Looks reasonable to me! 🚀

PR #584, opened 2024-02-07
Move Status functions to a dropdown

jeffdaley:


web/app/components/document/sidebar.hbs
```diff
@@ -20,15 +20,89 @@
...
-              this.isDraft
-              "!bg-color-palette-neutral-500  !text-white"
-            }}
-          />
+          {{#if this.isDraft}}
+            <Doc::Status
+              data-test-doc-status
+              @status={{@document.status}}
+              @hideProgress={{true}}
+              @type="inverted"
```
jeffdaley, line 28: The `inverted` type is very close to `neutral-500` styles we had before.

jfreda (approved):
🚀

PR #587, opened 2024-02-07
Fix bug when re-adding approvers

jfreda:
Re-adding a document approver wasn't making it correctly to the database (because the record's `deleted_at` was set) - this PR fixes the bug.


jeffdaley (approved):

jeffdaley:
you no-scoped it

PR #590, opened 2024-02-08
Improve (locked) document error handling

jeffdaley:
Refactors and tests document error states. Adds functionality to lock a document if the server returns a 423 on a task.


web/tests/acceptance/authenticated/document-test.ts
```diff
@@ -169,7 +174,6 @@ module("Acceptance | authenticated/document", function (hooks) {
     this.server.create("document", {
       objectID: 1,
       title: "Test Document",
-      status: "WIP",
```
jeffdaley, line 172: The deleted attributes are factory defaults

jfreda (approved):

PR #601, opened 2024-02-16
Add empty states to `/projects`

jeffdaley:
Adds empty states to the projects index screens.


web/app/routes/authenticated/projects/index.ts
```diff
@@ -20,6 +20,8 @@ export default class AuthenticatedProjectsIndexRoute extends Route {
       .fetch(`/api/${this.configSvc.config.api_version}/projects`)
       .then((response) => response?.json());
 
+    console.log("projects", projects);
+
```
jeffdaley, line 24:
```suggestion
```

PR #602, opened 2024-02-20
Fix search filtering

jeffdaley:
Fixes a regression causing filters to not appear on the `/results` route. Plus a little cleanup.

### DIFF
--- a/web/app/templates/authenticated/results.hbs
+++ b/web/app/templates/authenticated/results.hbs
@@ -1,5 +1,5 @@
 {{page-title "Search Results"}}
 
-<Header::Toolbar @facets={{@facets}} />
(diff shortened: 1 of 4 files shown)


web/app/templates/authenticated/results.hbs
```diff
@@ -1,5 +1,5 @@
 {{page-title "Search Results"}}
 
-<Header::Toolbar @facets={{@facets}} />
+<Header::Toolbar @facets={{this.model.facets}} />
```
jeffdaley, line 3: This is the primary change

jfreda (approved):
Can we add any tests so this regression doesn't pop up again?

PR #603, opened 2024-02-20
Improve filtering UX; Add tests for searching/filtering

jeffdaley:
Improves the logic of the facet dropdowns and how they react to filters being activated.

We currently use a mix of AND and OR conditions which allows for searches like "show me docs that are in-review AND from either Waypoint OR Terraform." We're able to support OR conditions by requesting facets and results separately: We grab them both on pageLoad but only the results are updated when filtering. The facet dropdowns and the active filters are totally separate.

This has downsides: The dropdown item counts are essentially useless after the first filter is applied. Whatever number was there on pageLoad is the same that'll be there with 10 filters on. This can easily lead to unexpected empty results, where even though you're clicking a link that says [there are 40 docs associated with this filter], you end up on an empty screen. 

This PR removes OR conditions. You can now only enable one filter per filter type. Once you select "RFC" you have to remove it to add "PRD." But the trade-off is that the dropdowns react to the scope. Now when you filter to product:Hermes the items in the Owner dropdown are its authors with accurate counts. You can no longer filter yourself into a dead end. You may not agree with the UI but the UX makes sense.

I do hope to support OR searches again, but the implementation issues will take time to work out. However, none of the changes here will make that work any more difficult.

AND, IMPORTANTLY,

Adds Mirage filtering coverage for filtering and searching. These tests would have caught the regression in #602 and will surely prevent future bugs if I have any say in it.

### DIFF
--- a/web/app/components/header/active-filter-list-item.hbs
+++ b/web/app/components/header/active-filter-list-item.hbs
@@ -1,4 +1,8 @@
-<LinkTo @query={{this.query}} class="active-filter-list-item">
+<LinkTo
+  data-test-active-filter-link
+  @query={{this.query}}
+  class="active-filter-list-item"
+>
   <FlightIcon @name="x" />
   {{@filter}}
 </LinkTo>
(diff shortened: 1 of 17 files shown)


jeffdaley:
Closing in favor of #605

Gonna table this UX until we can get it up to standard.

PR #605, opened 2024-02-21
Add Mirage coverage for filtering

jeffdaley:
Adds and tests Mirage coverage for searching and filtering.

### DIFF
--- a/web/app/components/results/index.hbs
+++ b/web/app/components/results/index.hbs
@@ -24,12 +24,25 @@
   </div>
 {{/if}}
 
-<h1 class="text-display-300 font-semibold">{{@results.nbHits}}
-  documents matching “{{@query}}”</h1>
+<h1
+  data-test-results-header-text
+  class="text-display-300 font-semibold"
+>{{@results.nbHits}}
+  {{if (eq 1 @results.nbHits) "document" "documents"}}
+  {{#if @query}}
+    matching
+    <span
+      class="bg-color-surface-warning text-color-foreground-warning-on-surface"
+    >
+      {{@query}}
+    </span>
+  {{/if}}
+</h1>
 <div class="flex w-full flex-col space-y-12 py-10">
   <div class="tile-list">
     {{#each @results.hits as |doc|}}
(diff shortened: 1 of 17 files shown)


web/app/components/results/index.hbs
```diff
@@ -24,12 +24,25 @@
...
+  class="text-display-300 font-semibold"
+>{{@results.nbHits}}
+  {{if (eq 1 @results.nbHits) "document" "documents"}}
+  {{#if @query}}
+    matching
+    <span
+      class="bg-color-surface-warning text-color-foreground-warning-on-surface"
+    >
+      {{@query}}
+    </span>
```
jeffdaley, line 38: This is styled more like our snippet highlight in the search popover.

web/app/services/algolia.ts
```diff
@@ -287,49 +287,38 @@ export default class AlgoliaService extends Service {
...
    */
-  getFacets = task(
-    async (
-      searchIndex: string,
-      params: AlgoliaSearchParams,
-      userIsOwner = false,
-    ): Promise<FacetRecords | undefined> => {
-      let query = params["q"] || "";
-      try {
-        let facetFilters = userIsOwner ? [`owners:${this.userEmail}`] : [];
```
jeffdaley, line 300: This only mattered when there were filters on the /my route. We'll bring this back if that's ever the case.

web/app/services/algolia.ts
```diff
@@ -287,49 +287,38 @@ export default class AlgoliaService extends Service {
...
-        let algoliaFacets = await this.searchIndex.perform(searchIndex, query, {
-          facetFilters: facetFilters,
-          facets: FACET_NAMES,
-          hitsPerPage: HITS_PER_PAGE,
-          maxValuesPerFacet: MAX_VALUES_PER_FACET,
-          page: params.page ? params.page - 1 : 0,
```
jeffdaley, line 306: Page precision doesn't matter; they all return the same facet results.

jfreda (approved):

PR #607, opened 2024-02-22
Redesign search results; Improve search UX

jeffdaley:
Redesigns the search results screen to use the `Doc::TileMedium` component. This allows us to show longer snippets at a more comfortable size, and it removes some `/results` specific-styles. (This change makes `Doc::Tile` obsolete.) In addition, there's a bunch of quality-of-life improvements:

- Loading a URL with `/results?q=foo` will now populate the search input
- Clicking the search input in the above scenario will trigger a search for that query
- Titles that match your query are now highlighted (popover and results screen)
- You no longer have to wait for the popover to load before being able to submit via enterKey
- Entering a new query now clears any previously active filters
- The `/results` page title now includes the query


pkg/algolia/client.go
```diff
@@ -139,10 +139,10 @@ func New(cfg *Config) (*Client, error) {
 
 		// Highlighting/snippeting
 		AttributesToSnippet: opt.AttributesToSnippet(
-			"content:7",
+			"content:30",
 		),
 		HighlightPostTag:    opt.HighlightPostTag("</mark>"),
-		HighlightPreTag:     opt.HighlightPreTag(`<mark class="hds-surface-warning hds-foreground-warning-on-surface">`),
```
jeffdaley, line 145: Handling this in CSS

web/app/components/results/index.hbs
```diff
@@ -1,61 +1,22 @@
...
-      <Hds::Link::Standalone
-        @text="View all {{this.capitalizedQuery}} documents"
-        @icon="arrow-right-circle"
-        @iconPosition="trailing"
-        @route="authenticated.documents"
-        @query={{hash product=(array this.capitalizedQuery)}}
-      />
-    </Hds::Card::Container>
-  </div>
-{{/if}}
```
jeffdaley, line 25: This isn't doing anything the search popover isn't already doing, and it takes up a ton of space. I have plans for this in the "All search results" view that will allow us to show all matches instead of 1.

jfreda (approved):

PR #639, opened 2024-03-07
Mirage cleanup (People route)

jeffdaley:
- Removes unused Mirage route (`/people`, `GET`)
- Moves `POST` route to a new section


jfreda (approved):
Oh glad to see we're not using the GET method for the people API. I think it's still supported, but we should remove it in the backend.

PR #646, opened 2024-03-13
Filter inaccessible docs out of Recently Viewed

jeffdaley:
Adds error logic to the RecentlyViewed service, fixing an issue where a deleted or private draft would break the dashboard.


jfreda (approved):
👍

PR #647, opened 2024-03-13
ModalAlertsService refactor

jeffdaley:
Converts string ModalTypes to enums. Refactors the `setActive` action to fix a flicker when hiding and immediately showing another modal.


jeffdaley:
Folding into another PR

PR #650, opened 2024-03-14
Sort application `<head>`

jeffdaley:
Sorts the application `<head>` based on recommendations from https://rviscomi.github.io/capo.js Why not


jfreda (approved):
I'm expecting a 500x improvement or I'll be disappointed.

PR #659, opened 2024-03-25
Search owner facet

jeffdaley:
Replaces the "Owner" dropdown with a search input. Now all owners are filterable, not just the top 100, and not just by email address (we now search the Google People API in addition to Algolia).

CleanShot 2024-03-25 at 11 24 18@2x

- Also adds a match count to the `/documents` route consistent with the `results?scope=documents` screen.
- Alphabetizes the dropdown lists and de-emphasizes the (often incorrect) count


jfreda (approved):
🚀

PR #667, opened 2024-04-02
Update FloatingUI package and add `is-dragging` logic

jeffdaley:
- Updates the FloatingUI package to get the `layoutShift` listener/property that allows us to react to changes in the layout.
- Adds logic to hide a popover when its parent contains the `.is-dragging` class, as can be the case in the forthcoming drag-and-drop UI.
- Resolves some Glint errors in the `content-test` file

### DIFF
--- a/web/app/components/floating-u-i/content.ts
+++ b/web/app/components/floating-u-i/content.ts
@@ -30,6 +30,7 @@ interface FloatingUIContentSignature {
     renderOut?: boolean;
     offset?: OffsetOptions;
     matchAnchorWidth?: MatchAnchorWidthOptions;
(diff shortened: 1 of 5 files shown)


web/app/components/floating-u-i/content.ts
```diff
@@ -65,7 +66,7 @@ export default class FloatingUIContent extends Component<FloatingUIContentSignat
   @action didInsert(e: HTMLElement) {
     this._content = e;
 
-    const { matchAnchorWidth, anchor, placement } = this.args;
```
jeffdaley, line 68: Unused variable

